PR #1, opened 2026-05-13
feat(interfaces): first-pass design for the three B-20 token variants

amiecorso:
## Summary

First-pass interface design for the three B-20 token variants: IDefaultToken (the inheritable base), IStablecoin, ISecurityToken, plus the Capabilities bitfield library that gates each variant's optional features.

Draft. Looking for feedback on the design choices documented in DESIGN_NOTES.md before iterating further.

## Design highlights

- Capabilities bitfield as the honest-signaling mechanism for permanently-disabled features (immutable per token, set at creation)
- Default IS Core: Stablecoin and Security extend IDefaultToken directly, no separate ICoreToken
- Compliance delegated to the policy engine; no internal blocklist
- Mint/burn role split, two-step admin transfer with delay (CCS-style)
- IStablecoin: per-minter rate limiting + ERC-3009 + currency identifier (informed by reading coinbase/custom-stablecoin)
- ISecurityToken: announce/share-ratio/create/redeem/admin-batch with on-chain announcement coupling

## Open for input

DESIGN_NOTES.md has a Design Rationale section explaining the why of the major decisions, plus a 9-item confirmation list at the bottom for the still-open questions. Most consequential of those: should we add a MEMOS_REQUIRED capability bit, and should adminBurn be allowed to affect any account (not just policy-blocked) given announcement coupling.

## Not implemented yet

- ITokenFactory.sol (one factory, three create methods, takes capability bitfield + initial supply per variant)
- IPolicyRegistry.sol (TIP-403 + TIP-1015 adapted)
- Reference Solidity implementations of all variants
- StdPrecompiles equivalent with per-variant address prefixes

### DIFF
--- a/src/interfaces/IDefaultToken.sol
+++ b/src/interfaces/IDefaultToken.sol
@@ -1,0 +1,313 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.20 <0.9.0;
+
+/// @title IDefaultToken
+/// @notice The base Solidity surface every Base-native token (B-20) implements.
+///         Variants (Stablecoin, Security, ...) extend this interface; nothing
+///         on this surface is variant-specific. A token created at the Default
+///         variant address presents exactly this interface.
+/// @dev    Backward-compatible with ERC-20 at the function-selector level:
+///         `transfer`, `transferFrom`, `approve`, `balanceOf`, `allowance`,
+///         `totalSupply`, `name`, `symbol`, `decimals` all match ERC-20 selectors.
+///         Memo'd siblings live alongside, and their existence does not change
+///         the ERC-20 selectors any wallet or contract already expects.
+///
+///         Every token's optional features are gated by an immutable
+///         `capabilities()` bitfield set at creation. Functions whose
+///         capability bit is unset revert with `FeatureDisabled` regardless
+///         of role state. See `Capabilities` for the bit definitions.
+interface IDefaultToken {
+    /*//////////////////////////////////////////////////////////////
+                                ERRORS
+    //////////////////////////////////////////////////////////////*/
+
+    error Unauthorized();
+    error ContractPaused();
+    error AlreadyPaused();
+    error NotPaused();
+    error InsufficientAllowance();
+    error InsufficientBalance(uint256 currentBalance, uint256 requestedAmount);
+    error InvalidAmount();
+    error InvalidRecipient();
+    error InvalidSupplyCap();
+    error SupplyCapExceeded();
+    error PolicyForbids();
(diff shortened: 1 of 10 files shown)


src/interfaces/ITokenFactory.sol
```diff
@@ -0,0 +1,271 @@
...
+    /*//////////////////////////////////////////////////////////////
+                                  TYPES
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Variant of a B-20 token. Recoverable from the token's
+    ///         address prefix; `NONE` indicates the address is not a B-20
+    ///         token created by this factory.
+    enum TokenVariant {
+        NONE,
+        DEFAULT,
```
ilikesymmetry, line 31: was assuming default and none are effectively the same, what's the case for separating you see?

src/interfaces/ITokenFactory.sol
```diff
@@ -0,0 +1,271 @@
...
+    ///                               Ignored when `initialSupply == 0`.
+    /// @param transferPolicyId       Initial value of `transferPolicyId`.
+    ///                               Must reference an existing policy in
+    ///                               the policy registry.
+    /// @param supplyCap              Initial value of `supplyCap`. Use
+    ///                               `type(uint256).max` for no cap.
+    /// @param contractURI            Initial ERC-7572 contract URI.
+    /// @param salt                   Caller-chosen salt for deterministic
+    ///                               address derivation.
+    struct CreateDefaultTokenParams {
```
ilikesymmetry, line 62: this is too big, needs to be smaller

ilikesymmetry: maybe a pattern we can consider is some way for factory smart contracts that wrap our factory precompile can also configure some of these things through the setter directly? Would help reduce our scope of creation params

src/interfaces/IDefaultToken.sol
```diff
@@ -0,0 +1,413 @@
...
+
+    /// @notice Sets the admin role for `role`. Requires `ADMIN_MUTABLE`
+    ///         capability and the current admin role for `role`.
+    /// @dev    REVERTS with `EnforcedDefaultAdminRules` when `role` is
+    ///         `DEFAULT_ADMIN_ROLE`; the default admin's admin role is
+    ///         always itself.
+    function setRoleAdmin(bytes32 role, bytes32 newAdminRole) external;
+
+    /*//////////////////////////////////////////////////////////////
+                          TWO-STEP DEFAULT ADMIN
```
ilikesymmetry, line 248: feeling resistance on including this although whenever we do admin patterns, I always advocate for 2-step patterns...

PR #7, opened 2026-05-16
Add PolicyRegistry implementation

eric-ships:
## Motivation

Implements `PolicyRegistry.sol` (TIP-403 + TIP-1015 parity) with a fourth compound slot for redeem policy specification. Bit-packing logic is isolated in `PolicySlot.sol`.

## Interface changes

- Added `redeemerPolicyId` as a fourth slot on `CompoundPolicyData`
- Updated `createCompoundPolicy` and `compoundPolicyData` signatures
- Added `isAuthorizedRedeemer` query
- Added `ALWAYS_REJECT` and `ALWAYS_ALLOW` to the `PolicyType` enum for built-in IDs

## Design decisions

**Single-slot storage.** Every policy fits in one `uint256`. Low 8 bits are always the type tag; remaining bits depend on type:
- WHITELIST/BLACKLIST: `[type(8b) | admin(160b)]` — 88 bits spare
- COMPOUND: `[type(8b) | sender(62b) | recipient(62b) | mintRecipient(62b) | redeemer(62b)]` — exactly 256 bits

Any authorization check costs at most 2 SLOADs: one for the compound slot, one for the constituent's member set.

**62-bit constituent IDs.** Four 64-bit IDs plus an 8-bit type tag would be 264 bits. Trimming 2 bits per ID makes it fit exactly. The interface keeps `uint64` throughout; truncation requires ~4.6e18 policies, which is not realistic.

**No recursion.** Constituents are validated as non-COMPOUND at creation time. The evaluation graph is always one level deep — no cycle detection or depth cap needed.

**Built-in IDs.** IDs 0 (always-reject) and 1 (always-allow) are handled as constants with no storage access. `ALWAYS_REJECT` and `ALWAYS_ALLOW` are added to the `PolicyType` enum so `policyData()` returns a meaningful type for built-ins rather than a misleading placeholder.

### DIFF
--- a/src/impls/PolicyDataLayout.sol
+++ b/src/impls/PolicyDataLayout.sol
@@ -1,0 +1,70 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.20 <0.9.0;
+
+import {IPolicyRegistry} from "../interfaces/IPolicyRegistry.sol";
+
+/// @title PolicyDataLayout
+/// @notice Internal library for encoding and decoding packed policy storage slots.
+///
+/// @dev Each policy is stored as a single uint256. The low 8 bits are always the
+///      PolicyType discriminator; the remaining bits depend on the type:
+///
+///        WHITELIST / BLACKLIST:
+///          [255:168] unused
+///          [167:8]   admin address (160 bits)
+///          [7:0]     PolicyType
+///
+///        COMPOUND:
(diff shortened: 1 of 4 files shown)


eric-ships:
Closing in favor of a new PR. The interface on main has been simplified significantly (no COMPOUND policies, spec naming, batch membership updates, ID 0 = always-allow). Reopening with a clean implementation that follows the current interface.

PR #23, opened 2026-05-19
feat(interfaces): sync IActivationRegistry to precompile implementation

eric-ships:
## Summary

The interface committed in #18 used `isActivated` / `admin` / `activate` (one-way semantics) but the underlying precompile implementation in [base/base#2733](https://github.com/base/base/pull/2733) uses `isEnabled` / `activationAdmin` / `enable` / `disable` (bidirectional). This PR aligns the Solidity surface to the actual ABI so generated selectors and error signatures match what the node dispatches.

Changes to `IActivationRegistry.sol`:

| Old | New |
|---|---|
| `isActivated(bytes32)` | `isEnabled(bytes32)` |
| `admin()` | `activationAdmin()` |
| `activate(bytes32)` | `enable(bytes32)` |
| (missing) | `disable(bytes32)` |
| `AlreadyActivated` | `AlreadyEnabled` |
| (missing) | `AlreadyDisabled` |
| `FeatureNotActivated` | `FeatureNotEnabled` |
| `FeatureActivated` event | `FeatureEnabled` event |
| (missing) | `FeatureDisabled` event |

Also adds `IActivationRegistry.sol` to the README file tree (it was missing from the listing added in #18).

## Open design question

base/base#2733's PR description says "enforces a one-way enable path" but the implementation ships `disable(bytes32)`. This interface mirrors the implementation. If the spec wins and `disable` is dropped from the precompile, this interface will need a follow-up that removes `disable`, `AlreadyDisabled`, and `FeatureDisabled` in the same hardfork PR. Flagging here so the decision lands before Beryl ships.

## Test plan

- `forge build` passes cleanly
- `StdPrecompiles.ACTIVATION_REGISTRY.isEnabled(...)` resolves to the correct 4-byte selector matching the precompile's dispatch table

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ forge install base/base-std
 src
 ├── <a href="./src/StdPrecompiles.sol">StdPrecompiles.sol</a>: Collection of precompiles and their interfaces
 └── interfaces
+    ├── <a href="./src/interfaces/IActivationRegistry.sol">IActivationRegistry.sol</a>: Runtime feature activation registry (Beryl+)
     ├── <a href="./src/interfaces/IB20.sol">IB20.sol</a>: Core Token Standard
     ├── <a href="./src/interfaces/IB20Stablecoin.sol">IB20Stablecoin.sol</a>: Stablecoin variant of B20
     ├── <a href="./src/interfaces/IB20Security.sol">IB20Security.sol</a>: Security variant of B20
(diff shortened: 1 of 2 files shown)


ilikesymmetry:
I think we want `activate/deactivate` language, not `enable/disable` per slack convos?

PR #26, opened 2026-05-19
Mock implementations and unit tests: `MockB20, `MockB20Stablecoin`, `MockTokenFactory`

amiecorso:
## Summary

Reference implementations for the `IB20` default-token surface, the
`IB20Stablecoin` variant, and the full `ITokenFactory.createToken`
flow. Lives under `test/lib/mocks/` and is etched via `vm.etch` per
Conner's mock-vs-src proposal. Written as Solidity-as-if-Rust:
spec-correspondence over Solidity idiom.

Out of scope:
- `IB20Security` (interface still in flux).
- `MockPolicyRegistry` / `MockActivationRegistry` (skeletons only for
  now; deferred to follow-ups).

### DIFF
--- a/src/interfaces/ITokenFactory.sol
+++ b/src/interfaces/ITokenFactory.sol
@@ -11,11 +11,14 @@ pragma solidity >=0.8.20 <0.9.0;
 ///         the initial admin.
 ///
 /// @dev    **Factory address.** The factory lives at the fixed address
-///         `0xB20...000F`.
+///         `0xB20F00000000000000000000000000000000000F`. The `0xB20F`
+///         prefix is deliberately distinct from the B-20 token prefix
+///         `0xB200` (byte `[1]` differs), so `isB20(factory)` returns
+///         false unambiguously.
 ///
 ///         **Token address schema (20 bytes).**
-///         - `[0:10]`  — `bytes10(0xB20...000)` shared prefix identifying a
-///                       factory-created B-20.
+///         - `[0:10]`  — `bytes10(0xB200000000000000000000)` shared
+///                       prefix identifying a factory-created B-20.
 ///         - `[10]`    — `bytes1(variant)` (matches `TokenVariant`).
 ///         - `[11]`    — `bytes1(decimals)` — encoded in the address so
 ///                       `decimals()` is recoverable statelessly from the
@@ -95,6 +98,15 @@ interface ITokenFactory {
     ///                      they execute with full privilege regardless
     ///                      of role state, and the privileged window
     ///                      closes the moment `createToken` returns.
+    ///                      May be `address(0)` for the "demonstrate no
+    ///                      owner" case (memecoins, credibly-neutral
+    ///                      tokens). When zero, no role is granted at
+    ///                      creation and the token has no admin: no
+    ///                      role grants, policy changes, or pauses are
+    ///                      ever possible post-creation. The
+    ///                      `renounceLastAdmin` path is the alternative
+    ///                      for tokens that need an admin during setup
+    ///                      and then want to evolve to admin-less.
     /// @param decimals      ERC-20 decimals. MUST be in `[2, 18]`.
     ///                      Encoded into address byte `[11]` for
     ///                      stateless retrieval.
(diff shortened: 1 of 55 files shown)


src/interfaces/ITokenFactory.sol
```diff
@@ -195,14 +204,21 @@ interface ITokenFactory {
...
+    ///         indicates the "demonstrate no owner" path. Post-creation
+    ///         role grants emit `RoleGranted` from the token as usual.
     event TokenCreated(
         address indexed token,
         TokenVariant indexed variant,
         string name,
         string symbol,
-        uint8 decimals
+        uint8 decimals,
+        address admin
```
ilikesymmetry, line 221: @stevieraykatz and I thought it'd be best to omit admin because we'll get a `RoleGranted` event for it from the token itself already.

test/lib/B20Test.sol
```diff
@@ -20,6 +20,28 @@ import {IB20} from "src/interfaces/IB20.sol";
 /// `unpauser`, `burnBlocker`) so role-gated tests have explicit named
 /// accounts to grant roles to in setUp's initCalls.
 contract B20Test is TokenFactoryTest {
+    // -- Role identifiers (match MockB20's keccak256 derivations) --
+    // Inlined here so tests can avoid `token.MINT_ROLE()` calls during
+    // setup paths where the token may not yet exist (e.g. createToken
+    // tests). Verified against MockB20's constants by direct equality
+    // in the role-constants test suite.
+    bytes32 internal constant DEFAULT_ADMIN_ROLE = bytes32(0);
```
ilikesymmetry, line 28: Should these be defined on the mock itself vs the test?

test/lib/mocks/MockB20Storage.sol
```diff
@@ -0,0 +1,175 @@
...
+        mapping(address owner => mapping(address spender => uint256 allowance)) allowances;
+        // ---------- Roles (OZ AccessControl-style) ----------
+        mapping(bytes32 role => mapping(address account => bool isMember)) roles;
+        mapping(bytes32 role => bytes32 adminRole) roleAdmins;
+        // Tracks DEFAULT_ADMIN_ROLE holder count so renounceRole can enforce
+        // LastAdminCannotRenounce in O(1). Bumped on grant, decremented on
+        // revoke / renounce.
+        uint256 adminCount;
+        // ---------- Policy slots ----------
+        mapping(bytes32 policyType => uint64 policyId) policyIds;
```
ilikesymmetry, line 59: Am realizing that we don't get clean packing into one slot for logical groups of policies (eg transfer sender/receiver/executor) if we use conventional mapping. Might need to be more custom than this

test/lib/mocks/MockTokenFactory.sol
```diff
@@ -1,37 +1,155 @@
...
+        //       slots the same way.
+        _writeBaseStorage(token, name_, symbol_, admin);
+        if (variant == TokenVariant.STABLECOIN) {
+            _writeStablecoinStorage(token, currency_);
+        }
+
+        // -- 6. Emit creation event BEFORE initCalls dispatch (per
+        //       ITokenFactory natspec) so init-call effects appear
+        //       strictly after the creation event in the log order.
+        //       Includes admin since there's no separate RoleGranted
```
ilikesymmetry, line 131: I guess it is net simpler to have one event to track like this. Alternative perspective is that there's only ever one event to look at for role changes if we force-emit `RoleGranted` from token on creation.

ilikesymmetry (commented):
Great progress with testing infra setup. I think we need to talk more about preferred layout for policies so that we can get 1 SLOAD for hot path of token transfer

src/interfaces/ITokenFactory.sol
amiecorso: so this was a point of debate with the AI... I think it depends how the rust implementation sets this -- whether it calls grantRole or just writes the slot. This might just be an artifact of what it meant to try to write the factory in the most rustlike way in solidity, but seems that calling the token grantRole might not happen

test/lib/mocks/MockB20Storage.sol
amiecorso: this is a good callout, will look closer tomorrow

test/lib/B20Test.sol
amiecorso: this might be redundant will look into

test/lib/mocks/MockTokenFactory.sol
amiecorso: yeah I think this might be the way: write the slot directly and force emit rolegranted from the token. there's probably a way to do this using foundry

PR #28, opened 2026-05-19
feat: IB20Security

stevieraykatz:
See https://docs.google.com/document/d/1rtEbvl8gJSw2SF3IoklaksdezzGpm_YzUr2We8BwTeQ/edit?tab=t.v42f4taogn6y


src/interfaces/IB20Security.sol
```diff
@@ -274,134 +163,102 @@ interface IB20Security is IB20 {
...
+    ///         from the inherited self-burn `burn(uint256)`; this
+    ///         batched form requires `BURN_ROLE` and operates on
+    ///         third-party balances for cold-path destruction.
+    ///
+    /// @dev    Requires `BURN_ROLE`. Subject to the `BURN` pause
+    ///         vector. Reverts on length mismatch or empty arrays.
+    ///
+    /// @param  accounts Accounts whose balances will be debited.
+    /// @param  amounts  Per-account amounts, parallel to `accounts`.
+    function burn(address[] calldata accounts, uint256[] calldata amounts) external;
```
ilikesymmetry: Curious if we should make these `batchMint` and `batchBurn` for clarity? And tbh, these could be in `IB20` core implementation too given generality

stevieraykatz: I could see them moving down for sure. calling them `batch..` is more a question for TANGOR since they need a more consistent ABI.

src/interfaces/IB20Security.sol
```diff
@@ -3,268 +3,157 @@ pragma solidity >=0.8.20 <0.9.0;
...
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice The current token-to-share ratio. A 1:1 ratio (numerator
-    ///         == denominator) means raw token balances equal share
-    ///         counts. A 2:1 ratio (e.g. after a 2-for-1 split) means
-    ///         each raw token represents 2 shares.
-    function shareRatio() external view returns (uint48 numerator, uint48 denominator);
+    /// @notice The current share-to-tokens ratio, scaled to the
+    ///         implementation's `WAD_PRECISION`.
+    function shareRatio() external view returns (uint128 sharesToTokensRatio);
```
ilikesymmetry: Noticing name diff between `shareRatio` and `sharesToTokenRatio`. Feels like it should just be one of them

src/interfaces/IB20Security.sol
```diff
@@ -3,268 +3,157 @@ pragma solidity >=0.8.20 <0.9.0;
...
     /// @notice The policy slot consulted against `msg.sender` on
-    ///         `redeem` / `redeemWithMemo`. Identifier is
-    ///         `keccak256("REDEEMER_SENDER")`. Security-specific because
-    ///         the redeem surface is itself security-specific; the
-    ///         underlying `policyId(bytes32)` mapping on `IB20` accepts
-    ///         any key, so this is a pure interface addition with no
-    ///         change to base storage shape.
+    ///         `redeem` and `redeemWithMemo`. Identifier is
+    ///         `keccak256("REDEEMER_SENDER")`.
     function REDEEMER_SENDER() external view returns (bytes32);
```
ilikesymmetry, line 126: Nit to solve in a separate PR, but I feel like this should be `REDEEMER_SENDER_POLICY` if we're at parity with `*_ROLE` format. Makes it a bit easier to tell from the constant what it's for.

ilikesymmetry (approved):
Huge checkpoint!

src/interfaces/IB20Security.sol
stevieraykatz: agreed. renaming.

src/interfaces/IB20Security.sol
stevieraykatz: Good callout. Let's do a separate pass for consistency on policies

src/interfaces/IB20Security.sol
stevieraykatz: Moved down to B20 and renamed to `batch...`

PR #30, opened 2026-05-20
Add CI: build, test, fmt, coverage

eric-ships:
## Motivation

No CI exists on the repo. With audit kickoff on May 25, every PR should be automatically gated on compilation, tests, and formatting so reviewers and auditors can trust the green checkmark.

## What changed

`.github/workflows/ci.yml` with three jobs that run on every push to main and every PR:

- **Build** (`forge build`): catches compilation errors; contract sizes printed as a separate informational step
- **Test** (`forge test -v`): runs the full test suite
- **Format** (`forge fmt --check`): prevents unformatted code from landing

All jobs use `foundry-rs/foundry-toolchain@v1` pinned to `stable` (currently v1.7.1).

How it was tested

Screenshot 2026-05-27 at 11 00 26 AM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,67 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - uses: foundry-rs/foundry-toolchain@v1
+
+      - name: Build
+        run: forge build --sizes
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - uses: foundry-rs/foundry-toolchain@v1
+
+      - name: Run tests
+        run: forge test -v
+
+  fmt:
+    name: Format
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - uses: foundry-rs/foundry-toolchain@v1
+
+      - name: Check formatting
+        run: forge fmt --check
+
+  coverage:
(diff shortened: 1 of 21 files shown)


refcell (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+          submodules: recursive
+
+      - uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: stable
+
+      - name: Run tests
+        run: forge test -v
+
+  fmt:
```
ilikesymmetry, line 42: not sure if CI runs in order of definition or parallelizes, but fmt should be before test if sequential given it will fail faster

test/lib/mocks/MockB20Storage.sol
```diff
@@ -407,9 +407,17 @@ library MockB20SecurityStorage {
     //                     TOP-LEVEL FIELD SLOTS
     // ============================================================
 
-    function sharesToTokensRatioSlot() internal pure returns (bytes32) { return slotOf(SHARES_TO_TOKENS_RATIO_OFFSET); }
-    function usedAnnouncementIdsBaseSlot() internal pure returns (bytes32) { return slotOf(USED_ANNOUNCEMENT_IDS_OFFSET); }
-    function identifiersBaseSlot() internal pure returns (bytes32) { return slotOf(IDENTIFIERS_OFFSET); }
+    function sharesToTokensRatioSlot() internal pure returns (bytes32) {
```
ilikesymmetry: I added ignore comments to keep these lean on other storage definitions. Mind adding those here and reverting the newlines? Same for the rest of the diff on this file

ilikesymmetry (approved):

.github/workflows/ci.yml
eric-ships: parallelizes

test/lib/mocks/MockB20Storage.sol
eric-ships: done - added `// forgefmt: disable-start` and `end`

PR #37, opened 2026-05-20
feat: Mock B20Security

stevieraykatz:
Adds a complete reference implementation for the IB20Security variant under test/lib/mocks/ and, in the process, evolves the interface itself most substantially around `announce`, which becomes the canonical "disclosure-and-execute" primitive instead of an unenforced indexer-pairing convention.

The interface changes are deliberate ABI/semantic shifts, not just docs: see the announce signature change, new errors, new role, redesigned batchBurn, and tightened redeem below. All changes should be back-ported to the spec doc by EoD.

### DIFF
--- a/src/interfaces/IB20Security.sol
+++ b/src/interfaces/IB20Security.sol
@@ -21,21 +21,24 @@ import {IB20} from "./IB20.sol";
 ///         these as-is and do not redeclare them here.
 ///
 ///         **Security-specific additions.** This interface adds:
-///         1. `announce(...)` plus `SECURITY_OPERATOR_ROLE` for posting
-///            holder-impacting disclosures (corporate actions, name
-///            changes, splits, etc.).
+///         1. `announce(bytes[],string,string,string)` plus
+///            `SECURITY_OPERATOR_ROLE`. The single canonical wrapper
+///            for posting a holder-impacting disclosure AND atomically
+///            executing the on-chain calls it discloses; see
+///            "Announcement pairing" below for the topology.
 ///         2. `sharesToTokensRatio()` / `toShares(...)` / `sharesOf(...)`
 ///            plus `updateShareRatio(...)` for split-safe
 ///            DeFi-compatible share accounting.
 ///         3. `batchMint(address[],uint256[])` and
 ///            `batchBurn(address[],uint256[])` for the cold-path
-///            corporate-actions issuance and seizure flows. These are
+///            corporate-actions issuance and clawback flows. These are
 ///            scoped to the security-token surface (not the base
 ///            `IB20`) because batched destruction of third-party
 ///            balances is a compliance-sensitive operation and the
-///            batched issuance path is paired with the announcement
-///            flow above. See the per-function natspec for role
-///            gating; `batchBurn` is held tighter than `batchMint`.
+///            batched issuance path is the natural target of the
+///            announcement bracket above. See the per-function
+///            natspec for role gating; `batchBurn` is held tighter
+///            than `batchMint`.
 ///         4. `redeem(...)` / `redeemWithMemo(...)` plus
 ///            `updateMinimumRedeemable(...)` and `minimumRedeemable()`
 ///            for the holder-initiated off-chain settlement path.
@@ -48,15 +51,45 @@ import {IB20} from "./IB20.sol";
 ///         tokens that want the corporate-actions desk to be the sole
 ///         caller of these functions grant `METADATA_ROLE` only to
 ///         addresses that also hold `SECURITY_OPERATOR_ROLE`. That
-///         pairing is operational, not contract-enforced.
+///         pairing is operational, not contract-enforced. The standard
+///         way to issue a name or symbol change is to wrap the
+///         `setName` / `setSymbol` call as an entry in
+///         `announce(...)`'s `internalCalls`, so the rebrand lands in
+///         the same `Announcement` ↔ `EndAnnouncement` bracket as the
+///         disclosure that explains it.
 ///
-///         **Announcement pairing.** The corporate-actions operator is
-///         expected to post an `announce(...)` alongside each
-///         state-changing operator call (`updateShareRatio`,
-///         `updateSecurityIdentifier`, `setName`, `setSymbol`) so that
-///         indexers can correlate the on-chain change with its
-///         off-chain disclosure. This interface does NOT enforce that
-///         pairing on-chain.
+///         **Announcement pairing.** Every state-changing operator
+///         call that affects holder-visible token semantics
+///         (`updateShareRatio`, `updateSecurityIdentifier`, `setName`,
+///         `setSymbol`, `batchMint`, `batchBurn`, and admin-level
+///         changes such as `updatePolicy` / `setSupplyCap` /
+///         `setContractURI` / `pause` / `unpause`) SHOULD be issued by
+///         encoding the call into the `internalCalls` parameter of
+///         `announce(...)`. The token then:
+///         1. emits `Announcement(caller, id, description, uri)`,
+///         2. `delegatecall`s each entry in `internalCalls` with
+///            `msg.sender` preserved (so the operator's own roles
+///            apply to the inner calls), reverting the entire
+///            announce if any inner call reverts, and
+///         3. emits `EndAnnouncement(id)`.
+///         This binds every change to its off-chain disclosure
+///         atomically: indexers never see a `Transfer`, a
+///         `ShareRatioUpdated`, or any other state-mutation event
+///         from a wrapped call without the surrounding bracket, and
+///         they never see a half-applied bracket because any inner
+///         revert unwinds the entire transaction including the
+///         `Announcement` event.
+///
+///         The bare functions remain individually callable by their
+///         role holders for emergency operator override, but
+///         unwrapped invocations produce no bracket events and so are
+///         indistinguishable from any other state mutation in the log
+///         stream. Production corporate-actions flows are expected to
+///         go through `announce(...)`; standalone calls are an escape
+///         hatch, not the standard path. Recursion (an inner call
+///         re-invoking `announce`) reverts with
+///         `AnnouncementInProgress` so the bracket is always exactly
+///         one level deep.
 interface IB20Security is IB20 {
     /*//////////////////////////////////////////////////////////////
                                  ERRORS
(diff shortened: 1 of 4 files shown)


src/interfaces/IB20Security.sol
```diff
@@ -101,6 +180,18 @@ interface IB20Security is IB20 {
...
+    ///         `internalCalls` has executed successfully (or
+    ///         immediately after `Announcement` itself for a pure
+    ///         announcement with `internalCalls.length == 0`).
+    ///         Carries the same `id` as the paired `Announcement` so
+    ///         indexers can join start ↔ end on the id even when
+    ///         scanning logs in isolation. The recursion guard
+    ///         (`AnnouncementInProgress`) makes the pairing
+    ///         within-tx unambiguous; the `id` field hardens cross-tx
+    ///         indexing as well.
+    event EndAnnouncement(string id);
```
ilikesymmetry, line 193: should we rename the first event `StartAnnouncement` for symmetry?

ilikesymmetry: Or rather `AnnouncementStarted` and `AnnouncementEnded`?

src/interfaces/IB20Security.sol
```diff
@@ -116,6 +207,29 @@ interface IB20Security is IB20 {
...
+                              PRECISION
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Fixed-point precision used to scale `sharesToTokensRatio`.
+    ///         Equal to `1e18` (one WAD). Exposed on the ABI so callers
+    ///         that read `sharesToTokensRatio()` directly can interpret
+    ///         the value without hardcoding the constant; typical
+    ///         callers should prefer `toShares(...)` / `sharesOf(...)`,
+    ///         which apply the precision internally.
+    function WAD_PRECISION() external view returns (uint256);
```
ilikesymmetry, line 231: I've honestly never grokked what "wad" means other than something like "a wad of cash is an amount of cash". Thoughts on something like `SHARE_RATIO_PRECISION`?

src/interfaces/IB20Security.sol
```diff
@@ -129,18 +243,78 @@ interface IB20Security is IB20 {
...
-    /// @notice Posts a holder-impacting announcement. Each `id` may be
+    /// @notice Posts a holder-impacting announcement and, in the same
+    ///         transaction, atomically executes the on-chain calls
+    ///         that the announcement describes. Each `id` may be
     ///         consumed at most once over the lifetime of the token;
     ///         subsequent calls that reuse `id` revert with
     ///         `AnnouncementIdAlreadyUsed`.
     ///
-    /// @dev    Requires `SECURITY_OPERATOR_ROLE`. Emits `Announcement`.
+    ///         Pass `internalCalls` as an empty array for a pure
```
ilikesymmetry, line 253: Not specific to this PR, but these natspec comments are getting out of hand in length. I think our AIs set a precedent of verbose in-lines that we are persisting, but this is an antipattern stemming from lack of root-level `docs/` or `spec/`.

test/lib/mocks/MockB20Security.sol
```diff
@@ -0,0 +1,300 @@
...
+///         on `redeem`) are NOT bypassed anywhere.
+contract MockB20Security is MockB20, IB20Security {
+    // ============================================================
+    //                          CONSTANTS
+    // ============================================================
+
+    bytes32 public constant SECURITY_OPERATOR_ROLE = keccak256("SECURITY_OPERATOR_ROLE");
+    bytes32 public constant BURN_FROM_ROLE = keccak256("BURN_FROM_ROLE");
+
+    bytes32 public constant REDEEMER_SENDER = keccak256("REDEEMER_SENDER");
```
ilikesymmetry: Flagging we should migrate policy type constants to `*_POLICY` format for parity with `*_ROLE`

test/lib/mocks/MockB20Security.sol
```diff
@@ -0,0 +1,300 @@
...
+    ///      call `announce` AND whatever role each inner function
+    ///      requires (e.g. `MINT_ROLE` for `batchMint`), so the
+    ///      authorization story is already enforced by the inner
+    ///      functions' own gates. The recursion guard exists to
+    ///      protect the EVENT topology (no nested brackets), not the
+    ///      authorization topology.
+    function _checkSelector(bytes calldata call) internal pure {
+        if (call.length < 4) revert InternalCallMalformed(call);
+        bytes4 sel = bytes4(call[:4]);
+        if (sel == ANNOUNCE_SELECTOR) revert AnnouncementInProgress();
```
ilikesymmetry: nit: feel like I'd rather in-line `IB20Security.announce.selector` to save me jumping up top to validate this selector matches expectation

test/lib/mocks/MockB20Security.sol
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    function sharesOf(address account) external view returns (uint256) {
+        return (MockB20Storage.layout().balances[account] * _sharesToTokensRatio()) / WAD_PRECISION;
+    }
+
+    function updateShareRatio(uint256 newSharesToTokensRatio) external {
+        _requireRole(SECURITY_OPERATOR_ROLE);
+        MockB20SecurityStorage.layout().sharesToTokensRatio = newSharesToTokensRatio;
+        emit ShareRatioUpdated(newSharesToTokensRatio);
```
ilikesymmetry, line 141: I really don't like the verbosity here, but feel slight smell not having terms 1:1 match:

```suggestion
    function updateSharesToTokensRatio(uint256 newSharesToTokensRatio) external {
        _requireRole(SECURITY_OPERATOR_ROLE);
        MockB20SecurityStorage.layout().sharesToTokensRatio = newSharesToTokensRatio;
        emit SharesToTokensRatioUpdated(newSharesToTokensRatio);
```

test/lib/mocks/MockB20Security.sol
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    // ============================================================
+    //                  BATCHED ISSUANCE / CLAWBACK
+    // ============================================================
+
+    function batchMint(address[] calldata recipients, uint256[] calldata amounts) external {
```
ilikesymmetry, line 148: missing `MINT_ROLE` check and not paused checks

test/lib/mocks/MockB20Security.sol
```diff
@@ -0,0 +1,300 @@
...
+            _mint(recipients[i], amounts[i]);
+        }
+    }
+
+    function batchBurn(address[] calldata accounts, uint256[] calldata amounts) external {
+        if (accounts.length != amounts.length) revert LengthMismatch(accounts.length, amounts.length);
+        if (accounts.length == 0) revert EmptyBatch();
+        if (!hasRole(BURN_FROM_ROLE, msg.sender)) {
+            revert AccessControlUnauthorizedAccount(msg.sender, BURN_FROM_ROLE);
+        }
```
ilikesymmetry, line 164: nit: we should probably just have an `onlyRole(ROLE)` modifier in shared B20 core if not already

test/lib/mocks/MockB20Security.sol
```diff
@@ -0,0 +1,300 @@
...
+
+    function securityIdentifier(string calldata identifierType) external view returns (string memory) {
+        return MockB20SecurityStorage.layout().identifiers[identifierType];
+    }
+
+    function updateSecurityIdentifier(string calldata identifierType, string calldata value) external {
+        _requireRole(SECURITY_OPERATOR_ROLE);
+        if (bytes(identifierType).length == 0) revert InvalidIdentifierType();
+        MockB20SecurityStorage.layout().identifiers[identifierType] = value;
+        emit IdentifierUpdated(identifierType, value);
```
ilikesymmetry: Full verbosity for 1:1 naming clarity with our two functions?
```suggestion
        emit SecurityIdentifierUpdated(identifierType, value);
```

test/lib/mocks/MockB20Security.sol
```diff
@@ -0,0 +1,300 @@
...
+    //                       POLICY OVERRIDES
+    // ============================================================
+
+    /// @dev Variant-first policy resolution: REDEEMER_SENDER lives in
+    ///      this variant's own packed slot; everything else falls
+    ///      through to the base. The base's UnsupportedPolicyType
+    ///      revert is the terminal case.
+    function _readPolicyId(bytes32 policyType) internal view virtual override returns (uint64) {
+        if (policyType == REDEEMER_SENDER) {
+            return uint64(MockB20SecurityStorage.layout().redeemPolicyIds);
```
ilikesymmetry: Callout we forgot to ping on in main channel: we think it's wise to move redeem storage out of `B20Security` namespace into a new separate `Redeemable` namespace in case we decide other variants (or core) should also have this functionality. Curious if that also intuits for you?

src/interfaces/IB20Security.sol
stevieraykatz: I don't think so. It contains the actual Announcement metadata and I think is appropriately named. `EndAnnouncement` is just a nice ergonomic lift for indexers.

src/interfaces/IB20Security.sol
stevieraykatz: fair point, but other standard interfaces have similarly verbose natspecs. See openzeppelin or solday for ref

src/interfaces/IB20Security.sol
stevieraykatz: `wad` comes from "wei amount denomination" from what I understand since its usually 1E18.

test/lib/mocks/MockB20Storage.sol
```diff
@@ -164,6 +164,101 @@ library MockB20Storage {
...
+        // Floor on `redeem` / `redeemWithMemo`, expressed in shares.
+        // Defaults to 0 (no floor); admin sets via updateMinimumRedeemable.
+        uint256 minimumRedeemable;
+        // ---------- Redeem-side policies (PACKED) ----------
+        // Layout:
+        //   [63:0]    redeemerSenderPolicyId
+        //   [127:64]  reserved
+        //   [191:128] reserved
+        //   [255:192] reserved
+        uint256 redeemPolicyIds;
```
ilikesymmetry: On storage alignment call, we discussed moving redeem functionality to its own `B20Redeemable` kind of mixin. For now, we only expect securities to implement this, but fundamentally a stablecoin use case could also be reasonable where a user burns stablecoin onchain to receive fiat offchain. Same thing if we consider other variants in the future that involve offchain reserves. 

The thought then is to proactively decouple integrating this feature from security storage space so it's easier to share across variants in the future (if needed). Does that feel good to you too?

test/lib/mocks/MockB20Security.sol
stevieraykatz: Will open a separate PR to fix this

test/lib/mocks/MockB20Security.sol
stevieraykatz: that's checked on `_mint`. Altenrative would be to hoist it up and introduce `_rawMint` like we do with `batchBurn`.

test/lib/mocks/MockB20Security.sol
stevieraykatz: happy to hoist since we would want to unify our storage pattern and interfaces accordingly. if we think we're going to hoist redeem though, intuition is to move all redemption-related methods and control to B20 layer.

test/lib/mocks/MockB20Security.sol
stevieraykatz: idk man, i think it's pretty clean as is.

test/lib/mocks/MockB20Security.sol
ilikesymmetry: Good to know. I think paused check is good to keep in `_mint`, but I'm more averse of including access control there. This is just a mock as well, so this is just a nit then if the functionality is baked in 👍

test/lib/mocks/MockB20Security.sol
stevieraykatz: will address in follow on PR

PR #49, opened 2026-05-21
chore: interface cleanup pass

ilikesymmetry:
## Summary

A grab-bag of interface cleanups across `ITokenFactory`, `IB20`, and `IB20Security`.

### `ITokenFactory`
- Remove `getTokenVariant` — variant byte is read directly off the address (`byte[10]`), no factory call needed.
- Add `isInitialized(address) -> bool` for the "factory bootstrap window closed" signal.
- Remove `NONE` from `TokenVariant`. The sentinel was only there to give `getTokenVariant` an "absent" return; now that the function is gone, `isB20` is the right predicate. Ordinals shift: `DEFAULT=0`, `STABLECOIN=1`, `SECURITY=2`.
- Document that the Security variant's `REDEEMER_SENDER_POLICY` slot defaults to ALWAYS_BLOCK; redemption must be opted into via `updatePolicy(REDEEMER_SENDER_POLICY, ...)` in `initCalls`.

### `IB20`
- Add `address indexed caller` as the first field of `Memo` so indexers can filter memo streams per account without a join to the preceding `Transfer`.
- Rename `set*` -> `update*` for consistency with the `*Updated` event convention:
  - `setName` -> `updateName`
  - `setSymbol` -> `updateSymbol`
  - `setSupplyCap` -> `updateSupplyCap`
  - `setContractURI` -> `updateContractURI`
  - `setRoleAdmin` -> `updateRoleAdmin`
- `RoleAdminChanged` deliberately **keeps** its OZ-AccessControl name (and exact signature) so OZ-aware indexers continue to decode it without a separate handler.

### Mocks & tests
- `MockB20` / `MockB20Security` / `MockTokenFactory` updated to match the new surface. `MockTokenFactory.isInitialized` reads the packed `initialized` byte directly via `vm.load` (no factory-only entrypoint on the token).
- Test files renamed (`setName.t.sol` -> `updateName.t.sol`, etc.); contracts and test functions renamed in lockstep.
- `getTokenVariant.t.sol` deleted. The `TokenVariant.NONE` revert test is replaced by an out-of-range-byte fuzz that exercises the same defensive branch via a low-level call.

## Test plan

- [x] `forge test` — 342 tests pass
- [x] `forge build` — clean
- [x] `forge fmt` — clean


test/lib/mocks/MockB20Storage.sol
```diff
@@ -171,20 +171,61 @@ library MockB20Storage {
     // the Rust impl validator) can read each field without remembering
     // the offset constant. Inlined by the compiler; zero runtime cost.
 
-    function nameSlot() internal pure returns (bytes32) { return slotOf(NAME_OFFSET); }
```
ilikesymmetry, line 174: @claude ideally we don't do this diff. I like the one-liner simplicity of the existing version. Is this a `forge fmt` issue?

ilikesymmetry: Yep, `forge fmt` artifact — it always expands single-statement function bodies and there's no config knob to opt out. Restored the one-liner block in 9eaeee7 and wrapped it in `// forgefmt: disable-start` / `// forgefmt: disable-end` so future `forge fmt` runs leave it alone. Verified by re-running `forge fmt` after the fix.

Heads-up for next time: the directive prefix is `forgefmt:` (no hyphen). The hyphenated form is silently ignored.

PR #65, opened 2026-05-22
feat: Add `name` to EIP-712 domain separator on B20Security

stevieraykatz:
A `name` change is a meaningful concept in the realm of Securities. Thus a name change explicitly changes the domain separator and invalidates previously signed permits. 

PR currently only shows changes to interface and mocks. Will update test if we get alignment that we should do this.

### DIFF
--- a/test/lib/mocks/MockB20Security.sol
+++ b/test/lib/mocks/MockB20Security.sol
@@ -246,6 +246,71 @@ contract MockB20Security is MockB20, IB20Security {
     }
 
     // ============================================================
+    //              EIP-712 DOMAIN OVERRIDE (ERC-5267)
+    // ============================================================
+
+    /// @dev Variant domain typehash: like the base typehash, but with
+    ///      `string name` as the leading field. The struct hash binds
+    ///      `keccak256(bytes(name))` (read live from storage on each
+    ///      call), so any successful `updateName(...)` invalidates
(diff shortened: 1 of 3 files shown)


test/lib/mocks/MockB20Security.sol
```diff
@@ -245,6 +245,71 @@ contract MockB20Security is MockB20, IB20Security {
...
+    ///      the `METADATA_ROLE` gate, the storage write, and the
+    ///      `NameUpdated` emission), then emit the ERC-5267
+    ///      `EIP712DomainChanged()` signal so off-chain integrators
+    ///      that cache `DOMAIN_SEPARATOR()` or `eip712Domain()`
+    ///      re-fetch. Event order is `NameUpdated` then
+    ///      `EIP712DomainChanged`, per the interface contract.
+    function updateName(string calldata newName) public override(MockB20, IB20) {
+        super.updateName(newName);
+        emit EIP712DomainChanged();
+    }
```
ilikesymmetry, line 311: seems like we should do this in B20 given we have permit domain there?

stevieraykatz: We don't ever update EIP712Domain in B20 so it didn't make obvious sense there

ilikesymmetry: but if we update name on B20, that would also update existing permits on normal B20?

PR #69, opened 2026-05-22
feat: add BatchSizeTooLarge error to IPolicyRegistry

rayyan224:
## Summary
- Adds `BatchSizeTooLarge(uint256 maxBatchSize)` to `IPolicyRegistry` to match the Rust PolicyRegistry precompile batch limit revert
- Documents that the error applies to `createPolicyWithAccounts`, `updateAllowlist`, and `updateBlocklist`

## Related
- base/base#2876

## Test plan
- [ ] Interface compiles with existing PolicyRegistry tests
- [ ] CI green

### DIFF
--- a/src/interfaces/IPolicyRegistry.sol
+++ b/src/interfaces/IPolicyRegistry.sol
@@ -92,6 +92,12 @@ interface IPolicyRegistry {
     /// @notice A required address argument was the zero address.
     error ZeroAddress();
 
+    /// @notice A membership batch exceeded the registry limit.
+    /// @param maxBatchSize The maximum number of accounts permitted per
+    ///        `createPolicyWithAccounts`, `updateAllowlist`, or
+    ///        `updateBlocklist` call.
+    error BatchSizeTooLarge(uint256 maxBatchSize);
+
     /// @notice `finalizeUpdateAdmin` was called for a policy with no
     ///         currently-staged pending admin.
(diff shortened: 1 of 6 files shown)


ilikesymmetry (commented):
if we have the error in interface, we should also include tests against our mocks so that we can validate intended behavior against real precompiles with fork tests

ilikesymmetry:
Added in `cb5cd92`. The mock now enforces the 64-account limit at the shared `_batchSetMembers` funnel and reverts with `BatchSizeTooLarge(MAX_BATCH_SIZE)`. Per-entrypoint coverage added:

- `test_createPolicyWithAccounts_revert_batchSizeTooLarge` + `_success_batchAtLimit`
- `test_updateAllowlist_revert_batchSizeTooLarge` + `_success_batchAtLimit`
- `test_updateBlocklist_revert_batchSizeTooLarge` + `_success_batchAtLimit`

The revert tests fuzz over arbitrary over-the-limit sizes; the success tests pin the inclusive boundary at exactly 64. `MAX_BATCH_SIZE` is kept as a compile-time literal on `PolicyRegistryTest` so the same tests are directly portable to fork-test mode against the real Rust precompile.

> Heads up: this branch was based on a fairly old `main` (predates several renames and the PolicyRegistry mocks refactor in #71). The new tests work on the current branch but will likely conflict on rebase — happy to rebase next.

PR #73, opened 2026-05-22
feat(factory): default REDEEM_SENDER_POLICY to ALWAYS_BLOCK_ID for security tokens

eric-ships:
[BOP-152](https://linear.app/coinbase/issue/BOP-152)

## Motivation

- Security tokens have redemption open by default because `REDEEM_SENDER_POLICY` is stored as a `uint64` lane in `redeemPolicyIds`, which defaults to zero (= `ALWAYS_ALLOW_ID`) from EVM storage.
- The spec requires redemption to be closed at creation: issuers must explicitly opt-in by pointing the slot at an allowlist policy, either atomically in `initCalls` or post-creation via `updatePolicy`.

## What changed

- `MockTokenFactory._writeSecurityStorage`: writes `ALWAYS_BLOCK_ID` into the `redeemPolicyIds` slot at creation time so the gate is closed before any `initCalls` run.
- `test/unit/TokenFactory/createToken.t.sol`: adds `test_createToken_success_securityDefaultsRedemptionToClosed` verifying both the ABI surface (`policyId(REDEEM_SENDER_POLICY)`) and the raw storage slot.
- `test/unit/B20Security/redeem/redeem.t.sol` and `redeemWithMemo.t.sol`: add `setUp` overrides that open redemption for success-path tests. Tests that verify the blocked state call `_setRedeemPolicy(ALWAYS_BLOCK_ID)` explicitly.
- `test/unit/B20Security/policy/policyId.t.sol`: updates the default assertion from `ALWAYS_ALLOW_ID` to `ALWAYS_BLOCK_ID`.
- `test/unit/B20Security/policy/updatePolicy.t.sol`: updates the slot-isolation assertion to reflect the new default.

441 tests pass.

- Relying on `initCalls` to close redemption: rejected because it would require every security token deployer to remember to include the `updatePolicy` call, creating an easy footgun for tokens shipped with redemption accidentally open.
- Closing redemption in `MockB20Security._redeemBurn` by hardcoding a check: rejected because the policy slot exists precisely to be configurable; the right fix is to seed it correctly at factory time.

### DIFF
--- a/src/StdPrecompiles.sol
+++ b/src/StdPrecompiles.sol
@@ -14,7 +14,7 @@ library StdPrecompiles {
     ///      disjoint (B-20 tokens have `0x00` at byte [1]; the factory has `0x0F`),
     ///      so `isB20(B20_FACTORY_ADDRESS)` returns false unambiguously. The
     ///      trailing `0x0F` byte echoes the prefix for visual symmetry.
-    address internal constant B20_FACTORY_ADDRESS = 0xB20F000000000000000000000000000000000000;
+    address internal constant B20_FACTORY_ADDRESS = 0xb20F00000000000000000000000000000000000f;
     address internal constant POLICY_REGISTRY_ADDRESS = 0x8453000000000000000000000000000000000002;
     address internal constant ACTIVATION_REGISTRY_ADDRESS = 0x8453000000000000000000000000000000000001;
 
--- a/src/StdPrecompiles.sol
+++ b/src/StdPrecompiles.sol
@@ -14,7 +14,7 @@ library StdPrecompiles {
     ///      disjoint (B-20 tokens have `0x00` at byte [1]; the factory has `0x0F`),
     ///      so `isB20(B20_FACTORY_ADDRESS)` returns false unambiguously. The
     ///      trailing `0x0F` byte echoes the prefix for visual symmetry.
-    address internal constant B20_FACTORY_ADDRESS = 0xB20f000000000000000000000000000000000000;
+    address internal constant B20_FACTORY_ADDRESS = 0xB20F000000000000000000000000000000000000;
     address internal constant POLICY_REGISTRY_ADDRESS = 0x8453000000000000000000000000000000000002;
     address internal constant ACTIVATION_REGISTRY_ADDRESS = 0x8453000000000000000000000000000000000001;
 
--- a/src/StdPrecompiles.sol
+++ b/src/StdPrecompiles.sol
@@ -14,7 +14,7 @@ library StdPrecompiles {
     ///      disjoint (B-20 tokens have `0x00` at byte [1]; the factory has `0x0F`),
     ///      so `isB20(B20_FACTORY_ADDRESS)` returns false unambiguously. The
     ///      trailing `0x0F` byte echoes the prefix for visual symmetry.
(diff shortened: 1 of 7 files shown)


src/StdPrecompiles.sol
```diff
@@ -14,7 +14,7 @@ library StdPrecompiles {
     ///      disjoint (B-20 tokens have `0x00` at byte [1]; the factory has `0x0F`),
     ///      so `isB20(B20_FACTORY_ADDRESS)` returns false unambiguously. The
     ///      trailing `0x0F` byte echoes the prefix for visual symmetry.
-    address internal constant B20_FACTORY_ADDRESS = 0xB20F000000000000000000000000000000000000;
+    address internal constant B20_FACTORY_ADDRESS = 0xb20F00000000000000000000000000000000000f;
```
ilikesymmetry: remove from this PR as we are aligned on previous address (no trailing `f`)

test/lib/mocks/MockB20Factory.sol
```diff
@@ -285,16 +286,20 @@ contract MockB20Factory is IB20Factory {
...
+    ///      it via `updatePolicy(REDEEM_SENDER_POLICY, <policyId>)` in
+    ///      `initCalls` or post-creation.
     function _writeSecurityStorage(address token, string memory isin_, uint256 minimumRedeemable_) internal {
         _writeString(token, MockB20SecurityStorage.identifierSlot("ISIN"), isin_);
         _writeUint(token, MockB20RedeemStorage.minimumRedeemableSlot(), minimumRedeemable_);
+        _writeUint(
+            token,
+            MockB20RedeemStorage.redeemPolicyIdsSlot(),
+            uint256(PolicyRegistryConstants.ALWAYS_BLOCK_ID)
+        );
```
ilikesymmetry, line 302: not sure if this is right, depends on how we're packing the slot.

ilikesymmetry: if we're at parity with how we pack slots, we have low bits empty which means we shouldn't do a simple `uint256` cast of the `uint64` value as this will occupy low bits

test/unit/B20Factory/createToken.t.sol
```diff
@@ -284,6 +285,26 @@ contract B20FactoryCreateB20Test is B20FactoryTest {
...
+            IB20Security(token).policyId(MockB20Security(token).REDEEM_SENDER_POLICY()),
+            PolicyRegistryConstants.ALWAYS_BLOCK_ID,
+            "REDEEM_SENDER_POLICY must be ALWAYS_BLOCK_ID at creation"
+        );
+        assertEq(
+            uint256(vm.load(token, MockB20RedeemStorage.redeemPolicyIdsSlot())),
+            uint256(PolicyRegistryConstants.ALWAYS_BLOCK_ID),
+            "redeemPolicyIds slot must hold ALWAYS_BLOCK_ID"
+        );
+    }
```
ilikesymmetry, line 306: not sure this is the right assert given packing comments above

eric-ships:
Closing in favour of #74 which was merged first and covers the same change. The only unique addition here was the factory-level cross-validation test (policyId() + raw slot dual assertion) which directly addresses Conner's packing concern — that can be added as a follow-on if desired.

PR #79, opened 2026-05-26
fix(factory): MissingRequiredField includes field name

eric-ships:
[BOP-155](https://linear.app/coinbase/issue/BOP-155)

## Motivation

`MissingRequiredField()` carried no information about which field was absent, making revert messages opaque in block explorers, Foundry traces, and error decoders. Callers had to read source to know what to fix.

## What changed

- `IB20Factory`: `error MissingRequiredField(string field)` — adds the field name to the error
- `MockB20Factory`: passes `"isin"` at the existing security-token check
- `MockB20Factory`: adds a missing empty-currency check for stablecoin tokens — an empty `currency` string previously passed the A–Z loop silently and was stored as an empty string; now reverts `MissingRequiredField("currency")`
- `createToken.t.sol`: updates `expectRevert` to encode the field name argument; adds `test_createB20_revert_missingCurrency` covering the new check

type=routine
risk=low
impact=sev5

### DIFF
--- a/src/interfaces/IB20Factory.sol
+++ b/src/interfaces/IB20Factory.sol
@@ -181,9 +181,9 @@ interface IB20Factory {
     ///         any known encoding for the requested variant.
     error UnsupportedVersion(uint8 version, B20Variant variant);
 
-    /// @notice A required string argument was the empty string (e.g.
-    ///         security `isin`).
-    error MissingRequiredField();
+    /// @notice A required string argument was the empty string.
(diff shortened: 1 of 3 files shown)


amiecorso:
Closing as duplicate of #95, which is the active version of this work and has been updated to align with the audit-response changes that landed on `main` in #89 (empty `currency` is now rejected via `InvalidCurrency("")`, separate from `MissingRequiredField`).

PR #81, opened 2026-05-26
feat: add b20 factory library

stevieraykatz:
Centralizes the encode-side of the `IB20Factory` surface in `src/lib/B20FactoryLib.sol` so integrators stop hand-rolling `abi.encodeCall` and parallel-array bundling per factory.

**Shape**
- Per-variant `params` encoders: `encodeDefaultCreateParams`, `encodeStablecoinCreateParams`, `encodeSecurityCreateParams`.
- Per-setter init-call encoders covering every init-callable method on `IB20` / `IB20Security` (supply cap, contract URI, name, policy, grant/revoke/setAdmin role, batchMint, identifier).
- Array builders for the common multi-entry shapes: `buildRoleGrants` (parallel `roles[]` / `accounts[]`, skips zero addresses), `buildSecurityIdentifierUpdates`, and a `concat` for stitching typed-core init calls with caller-supplied extras.
- `CREATE_PARAMS_VERSION = 1` as the single source of truth for the version byte; `LengthMismatch` lives on the library.

**Sweep-up**
- `B20Constants` gains `SECURITY_OPERATOR_ROLE` — it was the only security-applicable role still missing.
- `MockB20Factory` routes its three `p.version != 1` checks through `B20FactoryLib.CREATE_PARAMS_VERSION` so encode and decode agree by reference.
**Follow-up (not in this PR)**
- Library tests. Spike intentionally leaves them out.
- Downstream refactor: `tokenize/B20SecurityFactory` becomes a thin call-site over these helpers once it bumps `base-std`.


src/lib/B20FactoryLib.sol
```diff
@@ -0,0 +1,501 @@
...
+///         invokes on the freshly-deployed token during the privileged
+///         bootstrap window. Every `encode*` helper here produces
+///         exactly one such entry; `build*` helpers produce zero or
+///         more, in struct-field order.
+library B20FactoryLib {
+    /// @notice Current encoding version for every variant's
+    ///         `*CreateParams` struct. Carried as the leading `version`
+    ///         field so the factory can route between encodings as the
+    ///         schema evolves; today only `1` exists.
+    uint8 internal constant CREATE_PARAMS_VERSION = 1;
```
ilikesymmetry: We are expecting to version create params per variant so this might accidentally set assumption that there's one singular version for all variants? One alternative could be:
```solidity
uint8 internal constant B20_CREATE_PARAMS_VERSION = 1;
uint8 internal constant B20_STABLECOIN_CREATE_PARAMS_VERSION = 1;
uint8 internal constant B20_SECURITY_CREATE_PARAMS_VERSION = 1;
```

ilikesymmetry (approved):

PR #85, opened 2026-05-27
feat(factory): add StablecoinCreated and SecurityCreated companion events

eric-ships:
[BOP-158](https://linear.app/coinbase/issue/BOP-158)

## Motivation

Stablecoin `currency` and security `isin` are immutable post-creation with no setters, so they were unobservable from the event stream. Stream-based indexers (e.g. Coindexer) that cannot make mid-handler RPC calls had no way to index them.

## What changed

Two new events emitted immediately after `B20Created` in the same transaction:

- `StablecoinCreated(address indexed token, string currency)`
- `SecurityCreated(address indexed token, string isin)`

`B20Created` is unchanged. 10 new tests verify field values, ordering, and cross-variant isolation.

type=standard
risk=low
impact=sev5

### DIFF
--- a/src/interfaces/IB20Factory.sol
+++ b/src/interfaces/IB20Factory.sol
@@ -199,9 +199,9 @@ interface IB20Factory {
 
     /// @notice Emitted once per `createToken` invocation. The fields
     ///         cover the universal token-identity surface only;
-    ///         variant-specific state changes (e.g. `currency`, `isin`,
-    ///         supply cap, policy slots) are observable via the
-    ///         token's own events as they're applied during `initCalls`.
+    ///         variant-specific immutable fields (`currency`, `isin`) are
+    ///         announced in the companion events below, emitted in the same
+    ///         transaction immediately after `B20Created`.
     /// @dev    The initial admin grant (when `initialAdmin != address(0)`)
     ///         is announced via the standard `RoleGranted(DEFAULT_ADMIN_ROLE,
     ///         admin, factory)` event emitted from the token's own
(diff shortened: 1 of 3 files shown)


src/interfaces/IB20Factory.sol
```diff
@@ -213,6 +213,27 @@ interface IB20Factory {
...
+    event StablecoinCreated(address indexed token, string currency);
+
+    /// @notice Emitted immediately after `B20Created` when the new token is
+    ///         a Security variant. Carries the immutable `isin` field that is
+    ///         set at creation; additional identifiers (CUSIP, FIGI, SEDOL) can
+    ///         be attached post-creation via `IB20Security.updateSecurityIdentifier`.
+    /// @param token The address of the newly created security token.
+    ///              Matches the `token` field on the preceding `B20Created`.
+    /// @param isin  The ISIN identifier written at creation (e.g. `"US0378331005"`).
+    event SecurityCreated(address indexed token, string isin);
```
ilikesymmetry, line 235: I don't see why we need this event. We already emit when security identifiers are updated on the token

src/interfaces/IB20Factory.sol
```diff
@@ -213,6 +213,27 @@ interface IB20Factory {
...
+    ///         a Stablecoin variant. Carries the immutable `currency` field
+    ///         that is set at creation and has no setter on the token.
+    ///         Stream-based indexers (e.g. those that cannot make mid-handler
+    ///         view calls) should subscribe to this event to populate a
+    ///         `currency` column on their stablecoin token tables.
+    /// @param token    The address of the newly created stablecoin token.
+    ///                 Matches the `token` field on the preceding `B20Created`.
+    /// @param currency The ISO 4217 currency identifier (e.g. `"USD"`, `"EUR"`).
+    ///                 Uppercase ASCII letters (`A`–`Z`); immutable after creation.
+    event StablecoinCreated(address indexed token, string currency);
```
ilikesymmetry, line 226: Feels crufty. We only think we want this event in the factory because we don't have it on the token. We don't have it on the token because the currency is immutable. Adding a second event on the factory just for this field feels off. Would be curious to get your take on generalizing (of which I see a couple options, but none feel great either).

amiecorso:
Superseding with **Option A** per the meeting decision. Instead of two companion events (`StablecoinCreated`, `SecurityCreated`), we're adding a single `bytes variantParams` field to `B20Created` with a version-prefixed variant-specific payload, mirroring the existing `createParams` pattern.

| Variant | `variantParams` payload |
|---|---|
| `DEFAULT` | empty (`""`) |
| `STABLECOIN` | `abi.encode(B20StablecoinEventParams { version, currency })` |
| `SECURITY` | empty (`""`) — `isin` and `minimumRedeemable` are mutable and already surfaced via their own update events |

This design is forward-compatible (any variant can promote from empty to non-empty in a future version), keeps a single creation event, and mirrors the encoding shape already established by `B20*CreateParams`.

New implementation tickets:
- Solidity (base-std): [BOP-158](https://linear.app/coinbase/issue/BOP-158) — repurposed for Option A, assigned to Amie
- Rust (precompile): [BOP-216](https://linear.app/coinbase/issue/BOP-216) — assigned to Eric

Closing this PR; the Option A PR will land on a fresh branch. Thanks for the initial design work on companion events — that motivated the discussion that led to the chosen direction.

PR #89, opened 2026-05-28
audit: align mocks to Rust for claims 2-6 + parity tests

amiecorso:
## Summary

Response to the recent B20 audit pass. For each of the 6 specific discrepancies the auditor flagged, this PR either (a) brings the Solidity mock into selector-level parity with the Rust precompile and adds an explicit test that would catch divergence, or (b) adds the missing parity test where the behavior was already correct.

Verified against current `main` on both repos (base-std@4e6f474, base@f9d9d2804). All 506 unit tests pass mock-only and 504/0/2 in fork mode against the Rust precompile (the 2 skips are pre-existing unrelated slot-helper tests).

The auditor's two discussion points (check ordering, storage cleanup) are intentionally out of scope here. Check ordering is a separate effort (differential test harness across precondition pairs, planned as a follow-up PR). Storage cleanup was based on an incorrect premise — `mapping[k] = 0` and `delete mapping[k]` are byte-identical at the EVM trie level.

## Per-claim breakdown

| # | Claim | Source change | Test added |
|---|---|---|---|
| 2 | Empty stablecoin currency | Add `if (cb.length == 0) revert InvalidCurrency(...)` to `MockB20Factory.createB20` STABLECOIN branch. Fix `_isValidFiatCode("")` returning vacuously true. | `test_createB20_revert_emptyCurrency` |
| 3 | `getB20Address` invalid variant | None — ABI decoder rejects before any function body runs | `test_getB20Address_revert_outOfRangeVariant` |
| 4 | `batchBurn` zero amount | Add `if (amounts[i] == 0) revert InvalidAmount();` per-element in `MockB20Security.batchBurn` loop | `test_batchBurn_revert_zeroAmountInBatch` |
| 5 | `security_redeem_burn` zero amount | Add `if (amount == 0) revert InvalidAmount();` to `MockB20Security._redeemBurn` before shares math | `test_redeem_revert_zeroAmount` |
| 6 | `permit` malformed signature | None — already behavior-equivalent | `test_permit_revert_malformedSignature` |

Claim 1 (single vs three `createToken`) isn't a real divergence; both sides have a single parameterized dispatcher.

## Test results

- **Mock-only suite:** 506 passed (added 7 new tests)
- **Fork suite vs Rust f9d9d2804:** 504/0/2 (the 2 skips are pre-existing slot-helper tests, not introduced by this PR)

## Notes worth flagging

1. **Claim 2 has a `fixup!` commit on top.** The Rust precompile reverts `InvalidCurrency("")` on empty currency, not `MissingRequiredField()` as I initially guessed. The fork test surfaced it immediately. The fixup brings Solidity to the correct selector and renames the test from `_missingCurrency` to `_emptyCurrency`. Squash with the base commit if you prefer a single clean commit per claim.

2. **Claim 3 has no source change.** Solidity's ABI decoder rejects out-of-range enum bytes with `Panic(0x21)` before any function body runs; Rust rejects with `InvalidVariant()`. Both reject — different selectors. No way to make Solidity match without breaking the public ABI. New test asserts `vm.expectRevert()` (any revert) to pin the parity-of-rejection.

3. **Claim 6 has no source change.** Behavior was already equivalent — alloy returns `Err` on malformed sigs and maps to `InvalidSigner(0, owner)`, same as Solidity's defensive zero check. New test exercises the malformed `v=0` path as regression coverage.

4. **One side-effect bug fixed along the way:** `_isValidFiatCode("")` was returning vacuously true in the test helper, which is what made `vm.assume(!_isValidFiatCode(code))` silently exclude empty from every currency fuzz test. Fixed in claim 2's commit. Future fuzz tests of the same shape won't have this blind spot.

5. **Bound sweep completed for the audit-relevant surface.** Four success-test bounds relaxed from `[1, max]` to `[0, max]` on `amount`/`spendAmount` parameters in `allowance.t.sol`, `transferFrom.t.sol` (2 tests), and `transferFromWithMemo.t.sol`. The Rust precompile has no zero-amount guard on regular `mint`/`burn`/`transfer`/`transferFrom` (only `batchBurn` and `security_redeem` have them, both covered by claims 4 and 5), so the broader fuzz domain reproduces no new divergences against Rust — fork suite stays at 504/0/2. The other `[1, max]` sites in the suite are intentional: revert tests (bound is part of the setup), share-ratio tests (`ratio = 0` is mathematically degenerate), chainId tests (`chainId = 0` doesn't exist in the wild), and the redeem/batchBurn success tests where `[1, max]` is now the correct precondition after claims 4 and 5.

6. **Audit of remaining zero-edge fuzz exclusions.** Swept the rest of the suite for the same shape of bug. Two adjacent areas (empty value on `updateSecurityIdentifier`, stored zero on share-ratio reads) turned out to already be covered by existing tests at the base-function level (`emptyValueRemoves`, `defaultIsWad`, `zeroRestoresWadFallback`). Added two derived-function-level tests for the ratio fallback (`test_toShares_success_explicitZeroRatioFallsBackToWad`, `test_sharesOf_success_explicitZeroRatioFallsBackToWad`) so a refactor that reads the storage slot directly instead of going through `_sharesToTokensRatio()` would fail loudly. The remaining `bound(x, 1, max)` sites in the suite are intentional: revert tests, redeem/batchBurn success tests where `[1, max]` is the correct precondition after claims 4 and 5, and `chainId >= 1` in permit tests (chainId=0 doesn't exist in the wild).

## Out of scope (follow-ups)

- **Differential check-ordering test harness** (auditor's D1). Real divergence on `mint` (and likely `burnBlocked`). Design captured in cortex doc; planned as a separate PR.

## Commits

- `66f4afe` claim 2: reject empty stablecoin currency with MissingRequiredField
- `eaeb5fd` claim 3: add raw-bytes test for getB20Address out-of-range variant
- `3661d00` claim 4: reject zero amount per-element in batchBurn
- `84c4a84` claim 5: reject zero amount in security redeem with InvalidAmount
- `4106630` claim 6: parity test for malformed-signature permit recovery
- `9056a14` fixup! claim 2: align empty-currency selector to InvalidCurrency
- `2e334cd` fuzz: extend success-test amount bounds to cover zero on non-rejecting paths
- `70a6df1` gap 2: cover stored-zero share-ratio fallback at toShares + sharesOf level

### DIFF
--- a/test/lib/mocks/MockB20Factory.sol
+++ b/test/lib/mocks/MockB20Factory.sol
@@ -119,8 +119,12 @@ contract MockB20Factory is IB20Factory {
             if (p.version != B20FactoryLib.B20_STABLECOIN_CREATE_PARAMS_VERSION) {
                 revert UnsupportedVersion(p.version, variant);
             }
-            // Format check: every byte must be an uppercase ASCII letter (A-Z).
+            // Empty currency is a missing-required-field, not a format error: an empty byte
+            // string has no bytes to inspect so the format-check loop below would vacuously
+            // succeed without this explicit guard. Mirrors the SECURITY-variant isin check.
             bytes memory cb = bytes(p.currency);
+            if (cb.length == 0) revert MissingRequiredField();
+            // Format check: every byte must be an uppercase ASCII letter (A-Z).
             for (uint256 i = 0; i < cb.length; ++i) {
                 if (cb[i] < 0x41 || cb[i] > 0x5A) revert InvalidCurrency(p.currency);
             }
--- a/test/lib/mocks/MockB20Factory.sol
+++ b/test/lib/mocks/MockB20Factory.sol
@@ -119,11 +119,11 @@ contract MockB20Factory is IB20Factory {
             if (p.version != B20FactoryLib.B20_STABLECOIN_CREATE_PARAMS_VERSION) {
                 revert UnsupportedVersion(p.version, variant);
             }
-            // Empty currency is a missing-required-field, not a format error: an empty byte
-            // string has no bytes to inspect so the format-check loop below would vacuously
-            // succeed without this explicit guard. Mirrors the SECURITY-variant isin check.
+            // Empty currency must be rejected explicitly: the format-check loop below has
+            // no bytes to inspect on empty input and would vacuously succeed otherwise.
+            // Reverts InvalidCurrency("") to match the Rust precompile's selector.
             bytes memory cb = bytes(p.currency);
-            if (cb.length == 0) revert MissingRequiredField();
+            if (cb.length == 0) revert InvalidCurrency(p.currency);
             // Format check: every byte must be an uppercase ASCII letter (A-Z).
             for (uint256 i = 0; i < cb.length; ++i) {
                 if (cb[i] < 0x41 || cb[i] > 0x5A) revert InvalidCurrency(p.currency);
--- a/test/lib/mocks/MockB20Security.sol
+++ b/test/lib/mocks/MockB20Security.sol
@@ -199,6 +199,10 @@ contract MockB20Security is MockB20, IB20Security {
         if (accounts.length == 0) revert EmptyBatch();
         if (_isPaused(PausableFeature.BURN)) revert ContractPaused(PausableFeature.BURN);
         for (uint256 i = 0; i < accounts.length; i++) {
+            // Per-element zero-amount guard, matching the Rust precompile. Fires inside
+            // the loop before the balance check, so a zero in any slot reverts the whole
+            // batch (all-or-nothing atomicity).
+            if (amounts[i] == 0) revert InvalidAmount();
             _burnRaw(accounts[i], amounts[i]);
         }
     }
--- a/test/lib/mocks/MockB20Security.sol
+++ b/test/lib/mocks/MockB20Security.sol
@@ -199,10 +199,9 @@ contract MockB20Security is MockB20, IB20Security {
         if (accounts.length == 0) revert EmptyBatch();
         if (_isPaused(PausableFeature.BURN)) revert ContractPaused(PausableFeature.BURN);
         for (uint256 i = 0; i < accounts.length; i++) {
-            // Per-element zero-amount guard, matching the Rust precompile. Fires inside
-            // the loop before the balance check, so a zero in any slot reverts the whole
-            // batch (all-or-nothing atomicity).
-            if (amounts[i] == 0) revert InvalidAmount();
+            // Zero amounts are allowed per ERC-20 conventions (`transfer(0)` is valid);
+            // `_burnRaw` is a no-op for amount == 0 and emits `Transfer(account, 0, 0)`.
+            // Callers that want all-non-zero semantics can validate upstream.
             _burnRaw(accounts[i], amounts[i]);
         }
     }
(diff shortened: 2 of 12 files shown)


test/lib/mocks/MockB20Security.sol
```diff
@@ -199,6 +199,10 @@ contract MockB20Security is MockB20, IB20Security {
         if (accounts.length == 0) revert EmptyBatch();
         if (_isPaused(PausableFeature.BURN)) revert ContractPaused(PausableFeature.BURN);
         for (uint256 i = 0; i < accounts.length; i++) {
+            // Per-element zero-amount guard, matching the Rust precompile. Fires inside
+            // the loop before the balance check, so a zero in any slot reverts the whole
+            // batch (all-or-nothing atomicity).
+            if (amounts[i] == 0) revert InvalidAmount();
```
amiecorso: in B20 default, we allow 0 amount for operations (transfer requires 0 to be allowed per ERC20 spec, and we allow the same for burn and mint as they're forms of transfer and the consistency is nice / we should let people decide for themselves whether they allow 0 amounts which can be relevant to batch ergonomics etc.)

Do we want to allow 0 here after all? (same question below)

test/lib/mocks/MockB20Factory.sol
```diff
@@ -119,8 +119,12 @@ contract MockB20Factory is IB20Factory {
             if (p.version != B20FactoryLib.B20_STABLECOIN_CREATE_PARAMS_VERSION) {
                 revert UnsupportedVersion(p.version, variant);
             }
-            // Format check: every byte must be an uppercase ASCII letter (A-Z).
+            // Empty currency must be rejected explicitly: the format-check loop below has
+            // no bytes to inspect on empty input and would vacuously succeed otherwise.
+            // Reverts InvalidCurrency("") to match the Rust precompile's selector.
             bytes memory cb = bytes(p.currency);
+            if (cb.length == 0) revert InvalidCurrency(p.currency);
```
amiecorso, line 126: I think the only question here is whether we want this to revert with "InvalidCurrency" which is what the rust code does, or "MissingRequiredField"

amiecorso: Recommendation: stick with InvalidCurrency and match rust. more informative error.

test/lib/mocks/MockB20Security.sol
amiecorso: Defer to @stevieraykatz on how securities should function

stevieraykatz: "Allow 0" as a rule more closely follows ERC20 "treat 0-amount transfers as valid"

test/lib/mocks/MockB20Security.sol
```diff
@@ -292,6 +296,11 @@ contract MockB20Security is MockB20, IB20Security {
         if (!IPolicyRegistry(POLICY_REGISTRY).isAuthorized(REDEEMSenderPolicyId, msg.sender)) {
             revert PolicyForbids(REDEEM_SENDER_POLICY, REDEEMSenderPolicyId);
         }
+        // Explicit zero-amount guard matching the Rust precompile. Fires AFTER pause/policy
+        // and BEFORE the shares math, so zero-amount surfaces as InvalidAmount() rather than
+        // being absorbed by the shares == 0 path below (which is reserved for nonzero amounts
+        // that round to zero shares under low ratios).
+        if (amount == 0) revert InvalidAmount();
```
stevieraykatz: same logic as above. we should allow 0-burn

PR #98, opened 2026-05-29
fix(factory): take variant as uint8 to match Rust's InvalidVariant() selector (BOP-168)

amiecorso:
## Context

Tracking ticket: [BOP-168](https://linear.app/coinbase/issue/BOP-168)

Mirrors the Rust precompile's invalid-variant rejection on the Solidity reference. base-std currently diverges from the Rust precompile's selector on out-of-range variant bytes:

- **Rust** (`crates/common/precompiles/src/b20_factory/{storage,dispatch}.rs`): takes `u8` from alloy, calls `B20Variant::from_abi(call.variant).ok_or_else(... InvalidVariant {})`, reverts with `IB20Factory::InvalidVariant {}` selector.
- **Solidity** (pre-PR): takes the `B20Variant` enum, gets rejected by the ABI decoder with `Panic(0x21)` before any function body runs. The factory's `else { revert InvalidVariant(); }` branch is unreachable.

[base-std#89](https://github.com/base/base-std/pull/89) claim 3 noted this was *"no way to make Solidity match without breaking the public ABI"* — that note was incomplete: the wire-level 4-byte selector is identical for `(B20Variant,...)` and `(uint8,...)` forms (Solidity normalizes enum parameters to their underlying type at the canonical signature level), so no wire-ABI break is needed.

## Scope (expanded from ticket as written)

The ticket called out `getB20Address` but the same root cause affects `createB20`. Both Rust handlers use the identical `B20Variant::from_abi(...).ok_or_else(... InvalidVariant {})` pattern, both Solidity entry points previously had the same `Panic(0x21)` divergence, and both have parity tests (`test_createB20_revert_outOfRangeVariant`, `test_getB20Address_revert_outOfRangeVariant`) that previously asserted only "any revert." Migrating only one would leave an asymmetric interface. This PR fixes both.

## Fix

1. Change `IB20Factory.createB20` and `IB20Factory.getB20Address` signatures from `(B20Variant variant, ...)` to `(uint8 variant, ...)`.
2. Add entry-point guard `if (variant > uint8(type(B20Variant).max)) revert InvalidVariant();` to both `MockB20Factory.createB20` and `MockB20Factory.getB20Address`.
3. Inline `B20Variant(variant)` at usage sites inside `createB20` (binding to a local triggers stack-too-deep on Solc 0.8.30).
4. Update all in-tree callers to pass `uint8(IB20Factory.B20Variant.X)`.
5. Tighten both parity tests' `vm.expectRevert()` → `vm.expectRevert(IB20Factory.InvalidVariant.selector)`.

## Wire-ABI is unchanged

Verified via `forge inspect`:
```
| createB20(uint8,bytes32,bytes,bytes[]) | 62975e6a |
| getB20Address(uint8,address,bytes32)   | 8c30260f |
```
Both selectors are identical to the pre-PR enum-typed forms (Solidity's canonical signature lowers enums to their underlying `uint8`). External callers hitting the precompile by selector (TS clients, Rust precompile, raw calldata) are completely unaffected.

## Defense-in-depth

The existing `else { revert InvalidVariant(); }` branch in `createB20`'s dispatch chain is currently dead code (decoder kills out-of-range bytes first). After this PR the branch is live for a different reason: if a future variant is added to `B20Variant` without a corresponding dispatch arm in `createB20`, the new in-range byte falls through to this branch instead of being silently routed to `SECURITY` by the `else` covering it.

## Verification

- `forge build` clean
- Mock-only unit suite: **581 / 0 / 0**
- Targeted parity tests both pass with the tightened `InvalidVariant.selector` assertion:
  ```
  [PASS] test_createB20_revert_outOfRangeVariant(address,bytes32,uint8) (runs: 256)
  [PASS] test_getB20Address_revert_outOfRangeVariant(address,bytes32,uint8) (runs: 256)
  ```

## Diff stats

6 files changed, 104 insertions, 75 deletions. Most line changes are mechanical `uint8(IB20Factory.B20Variant.X)` casts at call sites in unit tests.

## Notes for review

- Inline `B20Variant(variant)` repeats ~10 times in `createB20`. Binding to a `B20Variant variant` local pushed the function over Solc's 16-local stack limit (LValue.cpp:54 stack-too-deep on the `_computeAddress` call). The inline-cast pattern is documented in the `createB20` NatSpec.
- `_computeAddress` keeps its `B20Variant` signature — it's internal, used only by the two entry points after they've already cast.
- `UnsupportedVersion(uint8, B20Variant)` error and `B20Created(..., B20Variant indexed, ...)` event keep their enum types — they're emitted *after* the variant has been validated and cast.

## Related

- Tracking: BOP-168
- Rust source of truth: `b20_factory::storage::B20FactoryStorage::create_b20` and `b20_factory::dispatch::dispatch::getB20Address` in `base/base`

### DIFF
--- a/src/interfaces/IB20Factory.sol
+++ b/src/interfaces/IB20Factory.sol
@@ -171,10 +171,14 @@ interface IB20Factory {
     ///         Caller must use a different salt.
     error TokenAlreadyExists(address token);
 
-    /// @notice `variant` is not a recognized `B20Variant`. Reached
-    ///         only when the factory is invoked with a raw variant byte
-    ///         outside the enum range (typed `createToken` callers are
-    ///         rejected by ABI decoding before this check fires).
+    /// @notice `variant` is not a recognized `B20Variant`. Raised when
+    ///         the factory is invoked with a `variant` byte outside the
+    ///         enum range (`> uint8(type(B20Variant).max)`). The variant
+    ///         parameter is typed as `uint8` (not the enum) on both
+    ///         `createB20` and `getB20Address` so the factory body runs
+    ///         this check before any decode-time panic, matching the
+    ///         Rust precompile's `IB20Factory::InvalidVariant {}` revert
+    ///         selector for every caller path (typed or raw-calldata).
     error InvalidVariant();
 
     /// @notice The leading `version` byte in `params` does not match
@@ -248,6 +252,12 @@ interface IB20Factory {
     ///         effects appear strictly after the creation event in the
     ///         log order.
     /// @param variant    Which variant struct `params` decodes as.
+    ///                   Typed as `uint8` (not the `B20Variant` enum)
+    ///                   so the factory body — not the ABI decoder —
+    ///                   rejects out-of-range bytes with the typed
+    ///                   `InvalidVariant()` revert, matching the Rust
+    ///                   precompile's selector. Solidity callers MUST
+    ///                   pass `uint8(B20Variant.X)`.
     /// @param salt       Caller-chosen salt for deterministic address
     ///                   derivation.
     /// @param params     ABI-encoded variant-specific creation struct,
@@ -258,7 +268,7 @@ interface IB20Factory {
     ///                   token-side authorization checks are bypassed
     ///                   for this window only.
     /// @return token     The address of the newly created token.
-    function createB20(B20Variant variant, bytes32 salt, bytes calldata params, bytes[] calldata initCalls)
+    function createB20(uint8 variant, bytes32 salt, bytes calldata params, bytes[] calldata initCalls)
         external
         returns (address token);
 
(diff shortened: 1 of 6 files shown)


amiecorso:
Closing — on further reflection, this PR trades real type-safety at the Solidity API surface (callers lose compile-time variant checking, every call site gets a permanent `uint8(B20Variant.X)` cast) for matching the Rust precompile's `InvalidVariant()` selector on an out-of-range error path that's already correctly rejected by Solidity's ABI decoder via `Panic(0x21)`.

The fork-test parity goal is real but the canonical answer here is arguably the reverse of the PR's framing: Solidity rejecting out-of-range enum bytes at the decoder is the correct Solidity behavior; Rust takes `u8` because Rust's ABI bindings don't have enum-typed inputs — that's a Rust-side constraint, not a Solidity-side requirement to match.

BOP-168 closed as won't-fix; the divergence can be addressed separately by documenting it in the mock and adjusting the two `test_*_revert_outOfRangeVariant` assertions if/when it becomes worth doing.

PR #99, opened 2026-05-29
chore(policy): PolicyRegistry Solidity-side parity cleanup (BOP-205, BOP-207, BOP-208)

amiecorso:
## Context

Two Roger-flagged audit cleanups against the Solidity `MockPolicyRegistry`:

- [BOP-207](https://linear.app/coinbase/issue/BOP-207) — validate batch size before creating policy
- [BOP-208](https://linear.app/coinbase/issue/BOP-208) — view function consistency cleanup

Both are Solidity-side parity work against the Rust precompile, both touch the same file (`MockPolicyRegistry.sol`), and both are small. Bundled into one PR with per-ticket sections below, mirroring [#89](https://github.com/base/base-std/pull/89)'s pattern.

> **Note:** BOP-205 (counter-exhaustion guard) was originally bundled here and has been dropped per review feedback — spec'ing a typed revert for an arithmetic boundary ~4.6 billion years out isn't worth the interface surface. The counter region of `_create()` reverts to its pre-PR state.

## Per-ticket breakdown

| # | Ticket | Source change | Test added |
|---|---|---|---|
| 1 | BOP-207 | Hoist `ZeroAddress` + `BatchSizeTooLarge` checks to the entry of `createPolicyWithAccounts` | `test_createPolicyWithAccounts_revert_zeroAdmin_precedes_batchSizeTooLarge` |
| 2 | BOP-208 | `policyExists` → use existing `policyExistsFromPacked` helper; `policyAdmin` comment; `pendingPolicyAdmin` adds built-in short-circuit | `test_pendingPolicyAdmin_success_zeroForBuiltinsEvenWithStoragePoison` (storage-poison defense-in-depth) |

## BOP-207 detail

Rust [`create_policy_with_accounts`](https://github.com/base/base/blob/main/crates/common/precompiles/src/policy/storage.rs) does:

```
validate_create_policy_inputs(admin, policy_type)  // catches ZeroAddress
require_account_batch_size(&accounts)              // catches BatchSizeTooLarge
create_policy_inner(...)
// write members
```

Solidity previously did `_create()` → `_batchSetMembers()` (batch check inside). After this PR:

```solidity
if (admin == address(0)) revert ZeroAddress();
if (accounts.length > MAX_BATCH_SIZE) revert BatchSizeTooLarge(MAX_BATCH_SIZE);
newPolicyId = _create(admin, policyType);
_batchSetMembers(...);
```

Downstream checks in `_create` / `_batchSetMembers` are kept (other entry points use them). The new test mirrors Rust's `create_policy_with_accounts_zero_admin_precedes_batch_size_revert` — when both fail, `ZeroAddress` wins.

## BOP-208 detail

Three small consistency changes in the view functions:

**`policyExists`**: switch from `packed != 0` to `MockPolicyRegistryStorage.policyExistsFromPacked(packed)`. Functionally equivalent given the encoding invariant; semantically precise; matches Rust's `packed.exists()`; survives any future encoding change that adds bits above the admin lane without setting the exists bit.

**`policyAdmin`**: comment-only. Documents why no explicit `exists()` gate is needed: the encoding invariant makes `_decodeAdmin(packed)` return `address(0)` for both never-written slots and intentionally-renounced slots, matching Rust's gated `(exists ? admin : 0)` observable.

**`pendingPolicyAdmin`**: add explicit built-in short-circuit. Without it, a corrupted `pendingAdmins[ALWAYS_ALLOW_ID]` slot would leak a non-zero address through the view. Mirrors Rust's `pending_policy_admin` gated read. Defense-in-depth, not a runtime bug.

The new test pokes `pendingAdmins[ALWAYS_ALLOW_ID]` and `pendingAdmins[ALWAYS_BLOCK_ID]` with `address(0xDEAD)` via `vm.store`, then asserts `pendingPolicyAdmin(builtin) == address(0)` for both. Fails before this PR (no short-circuit; the SLOAD returns the poison value), passes after.

## Verification

- `forge build` clean (only pre-existing warnings on unrelated files)
- Both new tests pass:
  ```
  [PASS] test_createPolicyWithAccounts_revert_zeroAdmin_precedes_batchSizeTooLarge(address,uint8,uint8) (runs: 256)
  [PASS] test_pendingPolicyAdmin_success_zeroForBuiltinsEvenWithStoragePoison() (gas: 8750)
  ```
- Full PolicyRegistry unit suite: **84 / 0 / 0**
- Mock-only full unit suite: **583 / 0 / 0** (was 581; exactly 2 new tests)

## Diff stats

3 files changed, 80 insertions, 3 deletions.

## Related

- BOP-207, BOP-208 (closed by this PR)
- Rust precompile reference: `crates/common/precompiles/src/policy/storage.rs` in `base/base`

### DIFF
--- a/test/lib/mocks/MockPolicyRegistry.sol
+++ b/test/lib/mocks/MockPolicyRegistry.sol
@@ -244,12 +244,8 @@ contract MockPolicyRegistry is IPolicyRegistry {
         _writeBuiltins();
         MockPolicyRegistryStorage.Layout storage $ = MockPolicyRegistryStorage.layout();
         uint56 counter = $.nextCounter;
-        // Counter-exhaustion guard. At one policy per 2-second block, hitting the
-        // 56-bit space (~7.2e16 values) takes ~4.6 billion years — so this branch
-        // is methodologically required, not practically reachable. Without the
-        // guard, an exhausted counter would wrap to 0 on the next increment and
-        // the next `_makeId` for a BLOCKLIST would collide with `ALWAYS_ALLOW_ID`.
-        if (counter == type(uint56).max) revert CounterExhausted();
+        // No overflow guard: at one policy per 2-second block, exhausting the
+        // 56-bit counter space (~7.2e16 values) takes ~4.6 billion years.
         unchecked {
             $.nextCounter = counter + 1;
         }
--- a/test/lib/mocks/MockPolicyRegistry.sol
+++ b/test/lib/mocks/MockPolicyRegistry.sol
@@ -92,6 +92,17 @@ contract MockPolicyRegistry is IPolicyRegistry {
         external
         returns (uint64 newPolicyId)
     {
+        // Match the Rust precompile's check precedence:
+        //   validate_create_policy_inputs (zero-admin) → require_account_batch_size →
+        //   create_policy_inner → write members
+        // Both checks are duplicated downstream (`_create` re-checks zero-admin for
+        // direct `createPolicy` callers, `_batchSetMembers` re-checks batch size for
+        // `updateAllowlist` / `updateBlocklist` callers). The hoisted entry-point
+        // copies ensure we revert before any `_create` mutation on the failing path
+        // AND pin the same revert-selector precedence Rust enforces (see Rust test
+        // `create_policy_with_accounts_zero_admin_precedes_batch_size_revert`).
+        if (admin == address(0)) revert ZeroAddress();
+        if (accounts.length > MAX_BATCH_SIZE) revert BatchSizeTooLarge(MAX_BATCH_SIZE);
         newPolicyId = _create(admin, policyType);
         _batchSetMembers({policyId: newPolicyId, policyType: policyType, value: true, accounts: accounts});
     }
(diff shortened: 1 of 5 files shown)


test/lib/mocks/MockPolicyRegistry.sol
```diff
@@ -213,8 +244,12 @@ contract MockPolicyRegistry is IPolicyRegistry {
         _writeBuiltins();
         MockPolicyRegistryStorage.Layout storage $ = MockPolicyRegistryStorage.layout();
         uint56 counter = $.nextCounter;
-        // No overflow guard: at one policy per 2-second block, exhausting the
-        // 56-bit counter space (~7.2e16 values) takes ~4.6 billion years.
+        // Counter-exhaustion guard. At one policy per 2-second block, hitting the
+        // 56-bit space (~7.2e16 values) takes ~4.6 billion years — so this branch
```
stevieraykatz: Are we really adding security checks for things that happen worst-case on the scale of billions of years? 
solidity: seems silly to spec this
rust: the built in uint64 under/overflow is probably more efficient than a custom handlder. 

Suggest "Won't Do" BOP-205.

amiecorso: agree

PR #102, opened 2026-05-29
feat(factory): add variantParams bytes field to B20Created event for stablecoin currency (BOP-158)

amiecorso:
## Context

Tracking ticket: [BOP-158](https://linear.app/coinbase/issue/BOP-158)

Adds a `bytes variantParams` field to `B20Created` so stream-based indexers (e.g. Coindexer) can recover the immutable `currency` of a stablecoin without making mid-handler RPC calls. The payload format mirrors the existing `createParams` pattern: ABI-encoded variant-specific struct prefixed by a version byte.

Supersedes the closed [#85](https://github.com/base/base-std/pull/85) (Option B: separate `StablecoinCreated`/`SecurityCreated` companion events). Rust precompile counterpart: [BOP-216](https://linear.app/coinbase/issue/BOP-216).

## Design

```solidity
event B20Created(
    address indexed token,
    B20Variant indexed variant,
    string name,
    string symbol,
    uint8 decimals,
    bytes variantParams  // NEW
);

struct B20StablecoinEventParams {
    uint8 version;
    string currency;
}
```

Indexer contract: variant comes from the indexed field; if `variantParams.length == 0`, no extra immutable fields; if non-empty, first byte is the version, decoder selected by `(variant, version)`. Same shape as `B20CreateParams` / `B20StablecoinCreateParams` / `B20SecurityCreateParams`.

## Per-variant payload

| Variant | `variantParams` payload | Rationale |
|---|---|---|
| `DEFAULT` | empty (`""`) | No extra immutable identity beyond name/symbol/decimals |
| `STABLECOIN` | `abi.encode(B20StablecoinEventParams { version: 1, currency })` | `currency` is immutable, no setter, no other event emits it |
| `SECURITY` | empty (`""`) | `isin` is mutable via `updateSecurityIdentifier` (emits `SecurityIdentifierUpdated`); `minimumRedeemable` is mutable via `updateMinimumRedeemable` (emits `MinimumRedeemableUpdated`) |

Forward-compatible: any variant can promote from empty to non-empty in a future version (just bump the variant's `*_EVENT_PARAMS_VERSION` and decode by `(variant, version)`).

## Why Option A over Option B (PR #85)

| Concern | Option A (this PR) | Option B (#85) |
|---|---|---|
| Indexer-friendly | ✅ one event per creation | ✅ but two events per creation |
| Variant-specific payload | ✅ typed via bytes + version | ✅ separate event per variant |
| Forward-compat for new immutable fields | ✅ bump version, add struct field | ❌ new event per change |
| Mirrors existing createParams pattern | ✅ same shape | ❌ asymmetric |
| Selector count growth as variants are added | ✅ stays 1 | ❌ grows linearly with variants |

## Changes

| File | What |
|---|---|
| `src/interfaces/IB20Factory.sol` | Add `bytes variantParams` to `B20Created`; add `struct B20StablecoinEventParams` |
| `src/lib/B20FactoryLib.sol` | Add `B20_STABLECOIN_EVENT_PARAMS_VERSION = 1` constant |
| `test/lib/mocks/MockB20Factory.sol` | Build `variantParams` per variant before `emit B20Created` |
| `test/unit/B20Factory/createToken.t.sol` | Update 2 existing event-match tests for the 6-arg signature; add `test_createB20_success_emitsB20Created_stablecoin` (expectEmit-level pin) and `test_createB20_success_b20CreatedVariantParams_stablecoin_decodes` (recorded-logs decode round-trip) |

## Verification

- `forge build` clean (only pre-existing warnings on unrelated files)
- Mock-only unit suite: **583 / 0 / 0** (stable across 3 consecutive runs; was 581 baseline + 2 new tests = 583)
- B20Factory suite: **55 / 0 / 0** (was 53 + 2 new = 55)
- All 4 `B20Created` emission tests pass:
  ```
  [PASS] test_createB20_success_emitsB20Created(address,bytes32) (runs: 256)
  [PASS] test_createB20_success_emitsB20Created_security(address,bytes32) (runs: 256)
  [PASS] test_createB20_success_emitsB20Created_stablecoin(address,bytes32) (runs: 256)
  [PASS] test_createB20_success_b20CreatedVariantParams_stablecoin_decodes(address,bytes32,uint256) (runs: 256)
  ```

## Diff stats

4 files changed, 170 insertions, 6 deletions.

## Notes for review

- Adds a `bytes memory variantParams` local at the bottom of `createB20`. Stayed under Solc's stack-too-deep limit; no inlining gymnastics needed (the local is allocated after `_writeBaseStorage` so several other locals are no longer live by the time it's introduced).
- The decode-level test (`test_createB20_success_b20CreatedVariantParams_stablecoin_decodes`) uses a deterministic seed → 3-letter uppercase generator (`_make3LetterUppercase`) rather than `vm.assume(_isValidFiatCode(...))` over a raw string. Avoids foundry's fuzz rejection-rate ceiling that the assume-filter approach hit on the first run.
- `_computeAddress` and `_writeBaseStorage` signatures are unchanged.

## Related

- Tracking: [BOP-158](https://linear.app/coinbase/issue/BOP-158)
- Rust counterpart: [BOP-216](https://linear.app/coinbase/issue/BOP-216)
- Parent decision: [CHAIN-4510](https://linear.app/coinbase/issue/CHAIN-4510)
- Superseded approach (Option B, companion events): [#85](https://github.com/base/base-std/pull/85) (closed)

### DIFF
--- a/src/interfaces/IB20Factory.sol
+++ b/src/interfaces/IB20Factory.sol
@@ -162,7 +162,7 @@ interface IB20Factory {
         uint256 minimumRedeemable;
     }
 
-    /// @notice Event payload carried in the `variantParams` field of
+    /// @notice Event payload carried in the `eventParams` field of
     ///         `B20Created` for STABLECOIN-variant tokens. ABI-encoded
     ///         and emitted with a leading `version` byte so stream-based
     ///         indexers can decode by `(variant, version)` without an
@@ -176,11 +176,11 @@ interface IB20Factory {
     ///                   `B20StablecoinCreateParams.currency`. Surfaced here
     ///                   because `currency` has no setter and no other
     ///                   event ever emits it; indexers that cannot make
-    ///                   mid-handler view calls (e.g. Coindexer) would
+    ///                   mid-handler view calls would
     ///                   otherwise have no way to recover it from the
     ///                   event stream.
     /// @dev    DEFAULT and SECURITY variants currently emit an empty
-    ///         `variantParams` byte string: DEFAULT has no extra
+    ///         `eventParams` byte string: DEFAULT has no extra
     ///         immutable identity fields beyond what's already in
     ///         `B20Created`, and SECURITY's `isin` and
     ///         `minimumRedeemable` are mutable and surfaced via their
@@ -247,7 +247,7 @@ interface IB20Factory {
     /// @param name          ERC-20 token name.
     /// @param symbol        ERC-20 token symbol.
     /// @param decimals      ERC-20 decimals (fixed per variant).
-    /// @param variantParams ABI-encoded variant-specific immutable identity
+    /// @param eventParams ABI-encoded variant-specific immutable identity
     ///                      fields, leading with a version byte. Empty
     ///                      (`""`) for DEFAULT and SECURITY (no extra
     ///                      immutable fields not already covered); for
@@ -255,12 +255,7 @@ interface IB20Factory {
     ///                      with `currency`. Indexers decode by
     ///                      `(variant, leading version byte)`.
     event B20Created(
-        address indexed token,
-        B20Variant indexed variant,
-        string name,
-        string symbol,
-        uint8 decimals,
-        bytes variantParams
+        address indexed token, B20Variant indexed variant, string name, string symbol, uint8 decimals, bytes eventParams
     );
 
     /*//////////////////////////////////////////////////////////////
--- a/src/interfaces/IB20Factory.sol
+++ b/src/interfaces/IB20Factory.sol
@@ -162,7 +162,7 @@ interface IB20Factory {
         uint256 minimumRedeemable;
     }
 
-    /// @notice Event payload carried in the `eventParams` field of
+    /// @notice Event payload carried in the `variantParams` field of
     ///         `B20Created` for STABLECOIN-variant tokens. ABI-encoded
     ///         and emitted with a leading `version` byte so stream-based
     ///         indexers can decode by `(variant, version)` without an
@@ -180,7 +180,7 @@ interface IB20Factory {
     ///                   otherwise have no way to recover it from the
     ///                   event stream.
     /// @dev    DEFAULT and SECURITY variants currently emit an empty
-    ///         `eventParams` byte string: DEFAULT has no extra
+    ///         `variantParams` byte string: DEFAULT has no extra
     ///         immutable identity fields beyond what's already in
     ///         `B20Created`, and SECURITY's `isin` and
     ///         `minimumRedeemable` are mutable and surfaced via their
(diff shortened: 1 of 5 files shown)


src/interfaces/IB20Factory.sol
```diff
@@ -162,6 +162,37 @@ interface IB20Factory {
...
+    /// @param  version   Event-encoding version. Currently `1`.
+    ///                   Independent of `B20StablecoinCreateParams.version`;
+    ///                   the event payload schema can evolve separately
+    ///                   from the create-call payload schema.
+    /// @param  currency  The immutable stablecoin currency identifier, same
+    ///                   value that was passed in
+    ///                   `B20StablecoinCreateParams.currency`. Surfaced here
+    ///                   because `currency` has no setter and no other
+    ///                   event ever emits it; indexers that cannot make
+    ///                   mid-handler view calls (e.g. Coindexer) would
```
stevieraykatz:
```suggestion
    ///                   mid-handler view calls would
```

src/interfaces/IB20Factory.sol
```diff
@@ -212,7 +242,26 @@ interface IB20Factory {
     ///         `B20Created` is the token-identity signal only. The
     ///         "demonstrate no owner" path (`initialAdmin == address(0)`)
     ///         skips the grant and emits no `RoleGranted` at bootstrap.
-    event B20Created(address indexed token, B20Variant indexed variant, string name, string symbol, uint8 decimals);
+    /// @param token         Address of the newly-created token.
+    /// @param variant       Which `B20Variant` was created.
+    /// @param name          ERC-20 token name.
+    /// @param symbol        ERC-20 token symbol.
+    /// @param decimals      ERC-20 decimals (fixed per variant).
+    /// @param variantParams ABI-encoded variant-specific immutable identity
```
stevieraykatz: Naming convention should match between `variantParams` and the underlying encoded data, i.e. `B20StablecoinEventParams.

src/interfaces/IB20Factory.sol
```diff
@@ -212,7 +242,21 @@ interface IB20Factory {
...
+    /// @param decimals      ERC-20 decimals (fixed per variant).
+    /// @param eventParams ABI-encoded variant-specific immutable identity
+    ///                      fields, leading with a version byte. Empty
+    ///                      (`""`) for DEFAULT and SECURITY (no extra
+    ///                      immutable fields not already covered); for
+    ///                      STABLECOIN, carries `abi.encode(B20StablecoinEventParams)`
+    ///                      with `currency`. Indexers decode by
+    ///                      `(variant, leading version byte)`.
+    event B20Created(
+        address indexed token, B20Variant indexed variant, string name, string symbol, uint8 decimals, bytes eventParams
```
ilikesymmetry: nit: 
```suggestion
        address indexed token, B20Variant indexed variant, string name, string symbol, uint8 decimals, bytes variantParams
```

PR #103, opened 2026-06-01
feat(ci): interface coverage check using forge coverage and lcov

eric-ships:

## Motivation

base-std tests are the behavioral spec for the base/base Rust precompile implementations. If a function exists in an interface but has no test coverage in the mock implementations, the Rust implementation can diverge silently.

## What changed

Adds `scripts/check-coverage.py` and a new `Interface Coverage` CI job that runs on every PR.

The script runs `forge coverage`, parses the resulting `lcov.info` for mock functions with zero hits, and exits 1 if any are found. No Solidity file parsing.

The CI job posts or updates a sticky PR comment on every commit showing either a green checkmark or the list of uncovered functions.

The `Interface Coverage` job runs independently of the `Forge Coverage` job even though both run `forge coverage`. Intentional: `Interface Coverage` is a blocking gate and `Forge Coverage` is informational. Coupling the gate to the informational job would mean a `Forge Coverage` failure silently skips the coverage check, and sequential execution adds latency. The extra parallel forge run is the right tradeoff.

Parsing interface `.sol` files directly to find missing test files (earlier version of this PR). Replaced after review feedback that it was fragile and reinvented a solved problem. The lcov-based approach uses forge's built-in coverage tooling as the source of truth.

Making `Interface Coverage` depend on the `Forge Coverage` artifact to avoid running `forge coverage` twice. Rejected because it couples a blocking check to an informational job, and adds artifact upload/download complexity with sequential latency.

type=routine
risk=low
impact=sev5

### DIFF
--- a/scripts/check-coverage.py
+++ b/scripts/check-coverage.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python3
+"""
+check-coverage.py
+
+Validates that every external function declared in base-std interface files
+has a corresponding unit test file. Exits 1 if gaps are found.
+
+Usage:
+  python3 scripts/check-coverage.py
+"""
+
+import re
+import sys
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parent.parent
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+# Maps interface filename -> unit test subdirectory under test/unit/
+INTERFACE_MAP = {
+    "IB20.sol":               "B20",
+    "IB20Security.sol":       "B20Security",
+    "IB20Stablecoin.sol":     "B20Stablecoin",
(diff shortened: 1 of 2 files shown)


scripts/check-coverage.py
stevieraykatz, line 1: There's a way to use the lcov report to do this without literally parsing the files in python. Not a fan of this approach; it seems fragile and a reinvention of a solved problem.

stevieraykatz (changes requested):

scripts/check-coverage.py
eric-ships: Good point. Rewrote the script to use forge coverage + lcov.info instead of parsing Solidity - runs forge coverage, then reads lcov.info for mock functions with 0 hits. No more regex over .sol files. The CI job now installs foundry before running it. See bc03b93.

stevieraykatz (approved):

PR #107, opened 2026-06-01
feat(ci): forge coverage job with per-file breakdown and PR comment

eric-ships:

## Motivation

base-std has one real Solidity library (`B20FactoryLib`) and a set of mock implementations that serve as the Rust precompile reference. Neither was measured for branch/line coverage. This adds a CI job that runs on every PR commit, posts a per-file coverage table as a sticky PR comment, and uploads an lcov artifact for local HTML browsing.

## What changed

Adds a `Forge Coverage` job to `.github/workflows/ci.yml`:
- Runs `forge coverage --no-match-coverage "(\.t\.sol|Test\.sol)\$"` to include mocks but exclude test runner files
- Generates `lcov.info` and uploads it as a CI artifact
- Posts/updates a per-file markdown table as a PR comment on every commit (using peter-evans/find-comment + create-or-update-comment so the comment updates in place)
- Informational only: does not block merging. A follow-on can add threshold enforcement once a stable baseline is visible.

Also pins `foundry-toolchain` to `v1.7.1` (matching `.foundry-version`) across all CI jobs to avoid GitHub API rate limit errors that occur when using `version: stable`.

Running `forge coverage` against only `src/` (excluding all of `test/`). Rejected because the mocks in `test/lib/mocks/` are the Solidity reference implementations used to validate the Rust precompile, so their branch coverage is exactly what we want to measure. The filter `(\.t\.sol|Test\.sol)\$` excludes test runner files while keeping mocks.

type=routine
risk=low
impact=sev5

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,9 +39,11 @@ jobs:
       - name: Run tests
         run: forge test -v
 
-  coverage:
+  forge-coverage:
     name: Forge Coverage
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
     steps:
       - uses: actions/checkout@v4
         with:
@@ -51,8 +53,18 @@ jobs:
         with:
           version: stable
 
-      - name: Generate coverage report
+      - name: Generate lcov report
         run: forge coverage --no-match-coverage "^test/" --report lcov
+
+      - name: Capture coverage summary
+        id: summary
+        run: |
+          output=$(forge coverage --no-match-coverage "^test/" 2>/dev/null | grep -E "^(\||File|Total)" || echo "No coverage data")
+          {
+            echo "output<<__EOF__"
+            echo "$output"
+            echo "__EOF__"
+          } >> "$GITHUB_OUTPUT"
 
       - name: Upload lcov report
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -34,11 +34,71 @@ jobs:
...
+          edit-mode: replace
+          body: |
+            <!-- forge-coverage -->
+            ### Forge Coverage (`src/lib/`)
+
+            ```
+            ${{ steps.summary.outputs.output }}
+            ```
+
+            Full lcov report: [download artifact](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}). To view locally: `genhtml lcov.info --branch-coverage -o coverage && open coverage/index.html`
```
stevieraykatz: The "To view locally" instructions don't include generating the coverage report in the first place. Consider wrapping in a makefile?

stevieraykatz: Downloading the file from Ci, extracting it, and then moving it into the repo and then running this works

eric-ships: Good call. Added a `Makefile` with a `make coverage` target that does the full flow (forge coverage + genhtml + open), and updated the PR comment to point at it. See 9fbff8b.

stevieraykatz (approved):

PR #115, opened 2026-06-01
refactor(factory): remove Default B20 variant (BOP-253)

ilikesymmetry:
## Summary

- Removes the `DEFAULT` value from `IB20Factory.B20Variant`; the enum now contains `STABLECOIN` (ordinal 0) and `SECURITY` (ordinal 1).
- Drops `B20CreateParams`, `B20FactoryLib.encodeDefaultCreateParams`, and the corresponding decode arm in `MockB20Factory`.
- Migrates the test base classes (`B20Test`, `B20StablecoinTest`) so generic IB20 tests now run against a stablecoin-variant token. The handful of tests that pinned default-variant identity (decimals=18, name=\"Test\", symbol=\"TST\") were retargeted to stablecoin equivalents.
- Updates `docs/B20/README.md` and `docs/B20/Factory.md` to drop the Default variant from the overview.

Linear: [BOP-253](https://linear.app/coinbase/issue/BOP-253/base-std-ib20factory-remove-default-variant). Spec-side counterpart to [BOP-242](https://linear.app/coinbase/issue/BOP-242/basebase-b20factory-remove-default-variant) (Rust precompile).

## Test plan

- [x] \`forge build\` clean (warnings only — pre-existing lints).
- [x] \`forge test\` — 600/600 pass.
- [ ] Coordinate with the Rust-side change on BOP-242 before this merges into a downstream consumer.


test/lib/B20Test.sol
```diff
@@ -58,18 +57,18 @@ contract B20Test is B20FactoryTest {
...
+    /// @dev    `MockB20Factory.createToken` etches `MockB20Stablecoin` runtime
     ///         bytecode at the computed address, writes initial state
     ///         directly via vm.store (no init function on the token),
     ///         runs initCalls, then closes the bootstrap window. Calls
     ///         against the returned `token` (transfer, mint, ...) execute
     ///         against a live mock token with the initial admin granted.
     function _deployToken() internal virtual returns (IB20) {
-        return IB20(_createDefault());
+        return IB20(_createStablecoin());
     }
```
ilikesymmetry, line 72: Are there any bad downstream effects of having a default `_deployToken` function use one of our variants?

src/interfaces/IB20Factory.sol
```diff
@@ -13,25 +13,11 @@ interface IB20Factory {
...
-        /// @dev Encoding version. Currently `1`.
-        uint8 version;
-        /// @dev ERC-20 token name.
-        string name;
-        /// @dev ERC-20 token symbol.
-        string symbol;
-        /// @dev Initial holder of `DEFAULT_ADMIN_ROLE`, or `address(0)` to deploy admin-less.
-        address initialAdmin;
+        SECURITY,
+        STABLECOIN
```
ilikesymmetry, line 20: intentional reordering given `B20Asset` is the more generic starting point for a `0` enum value with stablecoin as first special case

test/lib/mocks/MockB20.sol
```diff
@@ -48,7 +48,7 @@ import {B20Constants} from "src/lib/B20Constants.sol";
 ///         **What this mock is for:** local-tests-with-vm.etch and as a
 ///         readable spec artifact for auditors and Rust developers. It
 ///         lives under `test/` and is never deployed as production code.
-contract MockB20 is IB20 {
+abstract contract MockB20 is IB20 {
```
ilikesymmetry, line 51: making abstract to help enforce we can't deploy instances of this shared internal

test/unit/B20Factory/createToken.t.sol
```diff
@@ -224,37 +209,27 @@ contract B20FactoryCreateB20Test is B20FactoryTest {
...
 
-    /// @notice Verifies createToken returns the predicted address for the default variant
-    /// @dev Address determinism: returned address must equal getTokenAddress(DEFAULT, sender, salt)
-    function test_createB20_success_defaultMatchesPrediction(address caller, bytes32 salt) public {
-        _assumeValidCaller(caller);
-        IB20Factory.B20CreateParams memory p = _b20Params("Test", "TST", admin);
-        address predicted = factory.getB20Address(IB20Factory.B20Variant.DEFAULT, caller, salt);
-        address actual = _createDefault(caller, salt, p, new bytes[](0));
-        assertEq(actual, predicted, "createToken address must match prediction");
-    }
```
ilikesymmetry, line 256: Why was this deleted? use a variant to keep the test seems better?

test/unit/B20Factory/createToken.t.sol
```diff
@@ -461,27 +436,10 @@ contract B20FactoryCreateB20Test is B20FactoryTest {
...
-    ///      in the event).
-    function test_createB20_success_emitsB20Created(address caller, bytes32 salt) public {
-        _assumeValidCaller(caller);
-        IB20Factory.B20CreateParams memory p = _b20Params("MyToken", "MYT", admin);
-        address predicted = factory.getB20Address(IB20Factory.B20Variant.DEFAULT, caller, salt);
-
-        vm.expectEmit(true, true, false, true, address(factory));
-        emit IB20Factory.B20Created(predicted, IB20Factory.B20Variant.DEFAULT, "MyToken", "MYT", 18, bytes(""));
-        _createDefault(caller, salt, p, new bytes[](0));
-    }
```
ilikesymmetry, line 478: Why was this deleted? use a variant to keep the test seems better?

test/unit/B20Factory/createToken.t.sol
```diff
@@ -655,27 +613,6 @@ contract B20FactoryCreateB20Test is B20FactoryTest {
         IB20(token).mint(bob, 1);
     }
 
-    /// @notice Verifies createToken executes with admin == address(0) and grants no admin role
-    /// @dev "Demonstrate no owner" path: factory accepts zero admin, token has no admin afterward
-    ///      (no role grants, policy changes, or pauses ever possible). Replaces the prior
-    ///      _revert_zeroAdmin_default stub now that the design explicitly allows this.
-    function test_createB20_success_zeroAdminGrantsNoRole_default(address caller, bytes32 salt) public {
```
ilikesymmetry, line 662: Why was this deleted? use a variant to keep the test seems better?

test/unit/B20Factory/createToken.t.sol
ilikesymmetry: Deleted because it duplicated the variant-named coverage that exists on both arms after this PR:

- `test_createB20_success_stablecoinMatchesPrediction` (kept, lines ~242–252)
- `test_createB20_success_securityMatchesPrediction` (kept, lines ~254–264)

The deleted `_defaultMatchesPrediction` was the same assertion (returned address == `getB20Address(variant, sender, salt)`) on the now-removed DEFAULT arm. With Default gone, the same property is pinned per-variant by the two tests above. Happy to rename one of them to a generic `_matchesPrediction` if you'd prefer a single generic anchor — but with only two variants and both already covered, restoring a third copy felt like noise. Let me know if you want it back.

test/unit/B20Factory/createToken.t.sol
ilikesymmetry: Deleted because the per-variant emitters cover the same assertion shape on both arms:

- `test_createB20_success_emitsB20Created_security` (kept) — pins decimals=6, empty `variantEventParams`
- `test_createB20_success_emitsB20Created_stablecoin` (kept) — pins decimals=6 and the non-empty `B20StablecoinEventParams` payload

The deleted `_emitsB20Created` pinned decimals=18 + empty `variantEventParams` on the DEFAULT arm. With Default gone, every variant gets its own emitter test that pins its decimals and variantEventParams shape — there's no remaining "generic" emit behavior left to assert separately. Same offer as the other thread: I can rename one to drop the variant suffix if you want a generic-named anchor.

test/unit/B20Factory/createToken.t.sol
ilikesymmetry: Deleted because the zero-admin path is already pinned on both variants:

- `test_createB20_success_zeroAdminGrantsNoRole_stablecoin` (kept) — asserts no admin granted + adminCount=0 + initialized + currency slot preserved
- `test_createB20_success_zeroAdminGrantsNoRole_security` (kept) — same base assertions + ISIN slot preserved

The deleted `_default` version was the same zero-admin invariant against the DEFAULT arm. The two surviving tests have strictly stronger assertions (they also pin the variant-specific storage), so the deleted one wasn't adding coverage. Let me know if you want a rename to a generic anchor.

src/interfaces/IB20Factory.sol
eric-ships: ack

eric-ships (approved):
lgtm - appreciate the removal, then the name update

test/lib/mocks/MockB20Factory.sol
```diff
@@ -107,13 +107,17 @@ contract MockB20Factory is IB20Factory {
...
+        if (variant == B20Variant.SECURITY) {
+            B20SecurityCreateParams memory p = abi.decode(params, (B20SecurityCreateParams));
+            if (p.version != B20FactoryLib.B20_SECURITY_CREATE_PARAMS_VERSION) {
+                revert UnsupportedVersion(p.version, variant);
+            }
             name_ = p.name;
             symbol_ = p.symbol;
             admin = p.initialAdmin;
-            decimals = 18;
+            decimals = 6;
```
rayyan224, line 118: Assuming in a differnet mr we are gonna add in the asset changes ?

ilikesymmetry: Yep have separate tickets

PR #117, opened 2026-06-02
refactor(asset): remove batchBurn and BURN_FROM_ROLE (BOP-250)

ilikesymmetry:
## Summary

- Removes `batchBurn` from the security-variant interface, mock, and lib helpers; deletes its dedicated test directory.
- Removes `BURN_FROM_ROLE` from `B20Constants` plus all role-bundle / test references.

Linear: [BOP-250](https://linear.app/coinbase/issue/BOP-250/base-std-remove-batch-burn-method-and-burn-from-role). Counterpart to BOP-239 on the Rust precompile side.

## Test plan

- [x] `forge build` clean
- [x] `forge test` — full suite passes


docs/B20/Security.md
ilikesymmetry, line 1: ignore docs for now, will get big rewrite

test/lib/mocks/MockB20Security.sol
```diff
@@ -96,25 +94,6 @@ contract MockB20Security is MockB20, IB20Security {
...
-
-    /// @dev Like the base `onlyRole` but without the factory bootstrap
-    ///      bypass: the role check is unconditional, even for the
-    ///      factory in the init window. Reserved for variant paths
-    ///      that deliberately reject the bypass — currently just
-    ///      `batchBurn`, the corporate-actions clawback that has no
-    ///      init-time use case (see this contract's natspec). Lives
-    ///      here, not on the base, because no base function needs the
-    ///      stricter gate.
-    modifier onlyRoleStrict(bytes32 role) {
```
ilikesymmetry, line 111: why are we deleting this?

ilikesymmetry: Dead code after batchBurn removal: `onlyRoleStrict` had exactly one caller (`batchBurn` in `MockB20Security`), and its natspec called that out ("Reserved for variant paths that deliberately reject the bypass — currently just `batchBurn`"). With `batchBurn` gone there's no remaining caller and nothing in the security variant needs the strict gate, so I dropped the modifier rather than leaving it as an unused primitive. Easy to restore later if a new function actually needs the bypass-rejecting variant.

test/lib/B20Test.sol
```diff
@@ -141,13 +141,19 @@ contract B20Test is B20FactoryTest {
...
+    ///         `REDEEM_SENDER_POLICY`, because `_deployToken()` returns a
+    ///         security-variant token (see contract-level natspec) — fuzzing
+    ///         a `bytes32` that happens to equal `keccak256("REDEEM_SENDER_POLICY")`
+    ///         would otherwise fall through to the variant's `_readPolicyId`
+    ///         override and not revert as `UnsupportedPolicyType`.
     function _isKnownPolicyType(bytes32 policyType) internal pure returns (bool) {
         return policyType == B20Constants.TRANSFER_SENDER_POLICY || policyType == B20Constants.TRANSFER_RECEIVER_POLICY
-            || policyType == B20Constants.TRANSFER_EXECUTOR_POLICY || policyType == B20Constants.MINT_RECEIVER_POLICY;
+            || policyType == B20Constants.TRANSFER_EXECUTOR_POLICY || policyType == B20Constants.MINT_RECEIVER_POLICY
+            || policyType == keccak256("REDEEM_SENDER_POLICY");
```
ilikesymmetry, line 156: why are we adding this? Can we keep out as we are removing REDEEM_SENDER_POLICY in #119

ilikesymmetry: Keeping it in: without this line, this PR's CI breaks. The fuzz tests in question (test_policyId_revert_unsupportedPolicyType / test_updatePolicy_revert_unsupportedPolicyType) call vm.assume(!_isKnownPolicyType(x)) and then expect UnsupportedPolicyType. Because B20Test._deployToken() returns a security-variant token, REDEEM_SENDER_POLICY is actually a supported slot at runtime (via MockB20Security._readPolicyId override at test/lib/mocks/MockB20Security.sol:222) — when the fuzz seed hits keccak256("REDEEM_SENDER_POLICY") the call doesn't revert. This was a latent bug that the seed shift from the batchBurn removal happened to surface (see commit 91f1342).

You're right that it becomes dead the moment #119 lands — both the override and the constant go away — so removing this line is a trivial part of that PR's cleanup. Happy to leave a TODO here pointing at #119 if you'd like.

eric-ships (approved):

eric-ships (approved):

docs/B20/Security.md
stevieraykatz: can you link to where we're tracking this? seems like we'll need to take a fine-toothed anti-Security comb to the whole repo

stevieraykatz (approved):

PR #118, opened 2026-06-02
refactor(factory): remove isin and minimumRedeemable from security create params (BOP-254)

ilikesymmetry:
## Summary

Removes the `isin` and `minimumRedeemable` fields from `B20SecurityCreateParams` (the spec-side counterpart to BOP-243 on `base/base`). The factory no longer seeds `identifiers[\"ISIN\"]` or `minimumRedeemable` at creation; both surfaces are written exclusively post-creation (via `updateSecurityIdentifier` and `updateMinimumRedeemable` respectively, or default to the EVM zero of their slots).

Changes:
- `IB20Factory.B20SecurityCreateParams`: drop `isin` and `minimumRedeemable` fields.
- `IB20Factory.MissingRequiredField`: removed. ISIN was the only consumer; with that gone the error has no callers.
- `B20FactoryLib.encodeSecurityCreateParams`: signature collapses to `(name, symbol, initialAdmin)`.
- `MockB20Factory`: SECURITY decode arm stops reading `isin` / `minimumRedeemable`; `_writeSecurityStorage` no longer writes `identifiers[\"ISIN\"]` or `minimumRedeemable` (still seeds `REDEEM_SENDER_POLICY` to `ALWAYS_BLOCK_ID`).
- `B20FactoryTest._securityParams`: signature collapses to `(name, symbol, admin)`; `DEFAULT_ISIN` / `APPLE_ISIN` / `IDENTIFIER_ISIN` constants dropped (moved the test-fixture `IDENTIFIER_ISIN` to `B20SecurityTest` alongside `IDENTIFIER_CUSIP` / `IDENTIFIER_FIGI`).
- Tests pinning the ISIN-seeded-at-creation / minimumRedeemable-seeded-at-creation behavior are removed (`securitySeedsInitialState`, `securitySeededIsinEmitsNoEvent`, `returnsIsinSeededAtCreation`). `B20SecurityFullLayout` updated so the `identifiers[\"ISIN\"]` slot is asserted empty on a fresh token. `createB20_revertOrder.t.sol` drops its `version_beats_missingIsin` test (ISIN no longer validated at create time).
- `docs/B20/Security.md`: ISIN-at-creation framing removed; `minimumRedeemable` documented as defaulting to `0` at creation.

The redemption subsystem (`redeem`, `REDEEM_SENDER_POLICY`, `updateMinimumRedeemable`) is untouched — that's BOP-251's scope.

**Overlap note:** This PR overlaps with BOP-251 on the `minimumRedeemable` field. If BOP-251 lands first this PR may need a trivial rebase or partial no-op.

## Test plan

- [x] `forge build --force` clean
- [x] `forge test` — all 610 tests pass
- [ ] Reviewer sanity-check that the redemption surface (BOP-251 scope) is unchanged

### DIFF
--- a/docs/B20/Security.md
+++ b/docs/B20/Security.md
@@ -57,13 +57,13 @@ Redemptions let token holders initiate their own burn — typically the on-chain
 Two gates apply on every call:
 
 - `REDEEM_SENDER_POLICY` — the caller must be authorized.
-- `minimumRedeemable` — admin-set floor (in shares); redemptions below this floor revert with `BelowMinimumRedeemable`. Any redemption that resolves to zero shares (e.g. token dust against a large share ratio) is always rejected, even when `minimumRedeemable == 0`. Read with `minimumRedeemable()`; update with `updateMinimumRedeemable(newMin)` (admin-gated; emits `MinimumRedeemableUpdated`).
+- `minimumRedeemable` — admin-set floor (in shares); redemptions below this floor revert with `BelowMinimumRedeemable`. Any redemption that resolves to zero shares (e.g. token dust against a large share ratio) is always rejected, even when `minimumRedeemable == 0`. Defaults to `0` at creation. Read with `minimumRedeemable()`; update with `updateMinimumRedeemable(newMin)` (admin-gated; emits `MinimumRedeemableUpdated`).
 
 > ⚠️ **`REDEEM_SENDER_POLICY` defaults to `ALWAYS_BLOCK` at token creation** — redemptions are closed until the admin explicitly opens them via `updatePolicy`. The conservative default reflects that redemption is irreversible and regulator-sensitive.
 
 ## Security Identifiers
 
-Each Security token can carry one or more standardized identifiers (ISIN, CUSIP, FIGI, SEDOL, etc.). Read with `securityIdentifier(type)`; the value is a `string`. ISIN is required at creation via `B20SecurityCreateParams.isin`; other identifiers are optional and added post-creation.
+Each Security token can carry one or more standardized identifiers (ISIN, CUSIP, FIGI, SEDOL, etc.). Read with `securityIdentifier(type)`; the value is a `string`. All identifiers are optional and added post-creation — the factory does not seed any identifier at token creation.
 
 `updateSecurityIdentifier(type, value)` adds or updates an identifier and should be wrapped in `announce()`. Passing an empty `value` removes the entry. Unknown identifier types revert with `InvalidIdentifierType`.
 
--- a/src/interfaces/IB20Factory.sol
+++ b/src/interfaces/IB20Factory.sol
@@ -44,10 +44,6 @@ interface IB20Factory {
         string symbol;
         /// @dev Initial holder of `DEFAULT_ADMIN_ROLE`, or `address(0)` to deploy admin-less.
         address initialAdmin;
-        /// @dev International Securities Identification Number. Required; empty string reverts.
-        string isin;
-        /// @dev Initial value of `minimumRedeemable`. Use `0` to allow any non-zero redemption.
-        uint256 minimumRedeemable;
     }
 
     /// @notice Event payload carried in the `variantEventParams` field of `B20Created` for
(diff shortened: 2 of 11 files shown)


docs/B20/Security.md
ilikesymmetry, line 1: will ignore docs for now, going to be rewritten

src/interfaces/IB20Factory.sol
```diff
@@ -94,7 +99,7 @@ interface IB20Factory {
     ///         and before any `initCalls` are dispatched.
```
ilikesymmetry, line 94: don't remove error, we still need this for IB20Stablecoin missing currency

src/interfaces/IB20Factory.sol
ilikesymmetry, line 132: keep this, still need for IB20Stablecoin missing currency

test/unit/B20Factory/createToken.t.sol
ilikesymmetry, line 281: should we still keep this test but just assert that name, symbol, initialAdmin configured properly?

src/interfaces/IB20Factory.sol
ilikesymmetry: Addressed in c67d33a. Restored the `MissingRequiredField(string field)` error declaration and the matching `@dev Reverts with MissingRequiredField` line on `createB20`. Left the natspec example as `"currency"` since that's the live consumer the reviewer flagged; the mock still raises `InvalidCurrency("")` for empty currency, so the actual stablecoin switch can land in a follow-up without re-touching the interface.

src/interfaces/IB20Factory.sol
ilikesymmetry: Addressed in c67d33a (same commit as the error declaration thread above). Re-added the `/// @dev Reverts with MissingRequiredField when a required string field is empty.` line to `createB20`'s natspec.

test/unit/B20Factory/createToken.t.sol
ilikesymmetry: My read: with the seeded `ISIN` and `minimumRedeemable` gone there's no security-only initial state left for this test to pin — name/symbol/initialAdmin go through the same shared `_writeString` / role-grant path as DEFAULT and STABLECOIN, and they're already covered for the SECURITY variant by sibling tests in this file:

- `test_createB20_success_emitsB20Created_security` pins name/symbol via the `B20Created` event payload.
- `test_createB20_success_zeroAdminGrantsNoRole_security` exercises the admin path (zero-admin case); the non-zero case is implicit in every other security test that calls `_securityParams()` and later relies on `admin` holding `DEFAULT_ADMIN_ROLE`.
- `test_createB20_success_writesLongStrings` / `test_createB20_success_writesStringsAtEncodingBoundary` pin the storage encoding (shared across all variants since they share the writer).

There's no stablecoin equivalent of the old security-seeded-state test for the same reason (variant has no init state beyond `currency`, which is pinned in `test_createB20_success_zeroAdminGrantsNoRole_stablecoin` and the event tests). Happy to add a redundant security-only name/symbol/admin assertion if you'd prefer the symmetry — let me know.

eric-ships (approved):

test/lib/mocks/MockB20Factory.sol
```diff
@@ -161,7 +152,7 @@ contract MockB20Factory is IB20Factory {
         if (variant == B20Variant.STABLECOIN) {
             _writeStablecoinStorage(token, currency_);
         } else if (variant == B20Variant.SECURITY) {
-            _writeSecurityStorage(token, isin_, minimumRedeemable_);
+            _writeSecurityStorage(token);
```
stevieraykatz, line 155: Looks liek you're going to have a merge conflict with #119 from this.

test/lib/B20SecurityTest.sol
```diff
@@ -25,11 +25,13 @@ contract B20SecurityTest is B20Test {
...
-    // because the factory writes it during bootstrap too; CUSIP and
-    // FIGI are post-creation additions exercised only by the variant
-    // tests, so they belong on this base.
+    // Test-only identifier-type keys (`ISIN`, `CUSIP`, `FIGI`). All
+    // three are post-creation additions exercised only by the variant
+    // tests; the factory no longer seeds any identifier at bootstrap.
+
+    /// @notice Identifier-type key for the ISIN entry (International
+    ///         Securities Identification Number). Test-fixture only.
+    string internal constant IDENTIFIER_ISIN = "ISIN";
```
stevieraykatz, line 34: Should we even use an "ISIN" as a test param anymore? seems like we can get unit funcitonality without showing our hand that this was a security token

ilikesymmetry: Yep agree we shouldn't show this in general. Phasing this in rename PR when we transition away from "security identifier" naming for "custom metadata". Thought is to prioritize cutting first for minimal diffs on next for all the PRs coming though

PR #119, opened 2026-06-02
refactor(asset): remove redemption subsystem (BOP-251)

ilikesymmetry:
## Summary

Removes the holder-initiated redemption subsystem from `IB20Security` and its reference implementation. Counterpart to the base-side removal in BOP-240.

### Removed
- `IB20Security`: `redeem(uint256)`, `redeemWithMemo(uint256, bytes32)`, `minimumRedeemable()`, `updateMinimumRedeemable(uint256)`, `REDEEM_SENDER_POLICY()`, the `Redeemed` and `MinimumRedeemableUpdated` events, and the `BelowMinimumRedeemable` error.
- `IB20Factory.B20SecurityCreateParams.minimumRedeemable` field.
- `B20FactoryLib.encodeSecurityCreateParams` argument: now `(name, symbol, initialAdmin, isin)` only (signature change).
- `MockB20Security`: all redeem-side methods and the `_redeemBurn` helper.
- `MockB20Storage`: the `MockB20RedeemStorage` library (the entire `base.b20.redeem` ERC-7201 namespace, including `minimumRedeemable` and the `redeemPolicyIds` packed slot).
- `MockB20Factory._writeSecurityStorage`: no longer writes `minimumRedeemable` or seeds `REDEEM_SENDER_POLICY = ALWAYS_BLOCK_ID`; collapses to just the ISIN identifier write.
- Tests: `test/unit/B20Security/redeem/` (entire directory), `test/unit/B20Security/constants/policyTypeConstants.t.sol`, and redeem-specific cases inside `createToken.t.sol`, `policy/policyId.t.sol`, `policy/updatePolicy.t.sol`, `storage/B20SecurityFullLayout.t.sol`, `storage/MockB20Storage.t.sol`, and `B20FactoryLib/encodeSecurityCreateParams.t.sol`.
- Docs: the **Redemptions** section in `docs/B20/Security.md`, and the `REDEEM` mention in `docs/B20/README.md`'s pause section.

### Intentionally preserved
- **`IB20.PausableFeature.REDEEM` (ordinal `3`)** stays as a reserved enum value so existing serialized ordinals remain stable (the enum is documented as append-only). The reserved status is called out in the enum natspec and on `B20Constants.ALL_FEATURES_PAUSED`. No remaining operation consumes the bit, so pause/unpause tests that exercise it still pass — they assert pause-bookkeeping behavior only, not any gate.

## Test plan
- [x] `forge build` clean
- [x] `forge fmt` clean
[...]

PR #121, opened 2026-06-02
fix(test): include REDEEM_SENDER_POLICY in B20Test policy type helpers

eric-ships:
## Motivation

\`test_updatePolicy_revert_unsupportedPolicyType\` started failing on main after #115 changed \`B20Test._deployToken()\` to deploy a security-variant token. \`MockB20Security\` overrides \`_readPolicyId\`/\`_writePolicyId\` to handle \`REDEEM_SENDER_POLICY\`, so fuzz inputs that hit that scope no longer revert with \`UnsupportedPolicyType\`. But \`_isKnownPolicyType\` in \`B20Test\` only checked the 4 base-token policy types, leaving \`REDEEM_SENDER_POLICY\` in the "unsupported" partition — causing the test to assert a revert that no longer fires.

## What changed

Updated \`_knownPolicyType\` (mod 4 to mod 5, adds the fifth arm) and \`_isKnownPolicyType\` (adds the \`REDEEM_SENDER_POLICY\` check) in \`test/lib/B20Test.sol\` to reflect all 5 policy types supported by the security-variant token now deployed in setUp.

type=routine
risk=low
impact=sev5

### DIFF
--- a/test/lib/B20Test.sol
+++ b/test/lib/B20Test.sol
@@ -134,20 +134,22 @@ contract B20Test is B20FactoryTest {
     ///         extends the codomain (e.g. a Security test that also
     ///         covers `REDEEM_SENDER_POLICY`).
     function _knownPolicyType(uint8 idx) internal pure returns (bytes32) {
-        uint8 i = idx % 4;
+        uint8 i = idx % 5;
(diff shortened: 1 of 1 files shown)


ilikesymmetry:
REDEEM_SENDER_POLICY is getting removed here: https://github.com/base/base-std/pull/119

PR #125, opened 2026-06-02
feat(asset): add configurable decimals to security variant (BOP-252, BOP-255, BOP-259)

ilikesymmetry:
## Summary

Adds a configurable `decimals` field to the B-20 **security** variant. Today the security mock hardcodes `decimals()` to `6`; this change makes it a per-token, immutable value chosen at creation in the inclusive range `[6, 18]`. Stablecoin variant is unchanged (still hardcoded `6`).

Implements:
- [BOP-252](https://linear.app/coinbase/issue/BOP-252) — custom decimals storage on B20 asset
- [BOP-255](https://linear.app/coinbase/issue/BOP-255) — `decimals` on create params + event threading
- [BOP-259](https://linear.app/coinbase/issue/BOP-259) — test coverage

## Storage layout

Append-only addition to the `base.b20.security` ERC-7201 namespace:

```solidity
struct Layout {
    uint256 sharesToTokensRatio;  // slot 0 (unchanged)
    uint8   decimals;             // slot 1 (NEW; uint8 lands alone in this slot)
    mapping(string => bool)   usedAnnouncementIds;  // slot 2 (was 1)
    mapping(string => string) identifiers;          // slot 3 (was 2)
}
```

The two mapping offsets shifted by one (1 → 2, 2 → 3). Since both are pure mappings — their declared slot is the base used in `keccak256(abi.encode(key, baseSlot))` to derive value slots — the on-chain key derivations move with the offsets, no orphaned data. Pre-PR no security tokens exist outside tests, so this is layout-safe in practice as well.

A `decimalsSlot()` helper is added next to `sharesToTokensRatioSlot()`. `B20SecurityFullLayout.t.sol` is extended with the new slot assertion.

## ABI changes

- `IB20Factory.B20SecurityCreateParams` gains `uint8 decimals` after `initialAdmin`.
- `B20FactoryLib.B20_SECURITY_CREATE_PARAMS_VERSION` bumps `1 → 2`.
- `B20FactoryLib.encodeSecurityCreateParams(...)` signature gains the new `decimals` argument.
- New error `IB20Factory.InvalidDecimals(uint8 decimals)` thrown when `decimals` is out of range.
- `B20Created` event signature is unchanged; the existing `uint8 decimals` field now reflects the input value instead of being hardcoded `6` for security.
- `B20Constants` gains `MIN_ASSET_DECIMALS = 6` and `MAX_ASSET_DECIMALS = 18` (placed on `B20Constants` rather than `IB20Security` to match the existing pattern for plain validation bounds).

## Range rationale: `[6, 18]`

- `6` — the stablecoin / securities-platform floor; below it, integrations that quote unit prices in micro-units lose meaningful precision.
- `18` — the ERC-20 community ceiling that every wallet and indexer renders correctly; going higher risks integration breakage.

## Stablecoin variant unchanged

Stablecoin still hardcodes `decimals()` to `6` and ignores any per-create-params decimals field (there is no such field on `B20StablecoinCreateParams`). A regression test pins this behavior.

## Test plan

- [x] `forge build` clean
- [x] `forge fmt` clean (with the known `B20FactoryLibTest.sol` formatting quirk discarded)
- [x] Full suite: `forge test` → 578 tests pass
- [x] New file at `--fuzz-runs 1000`: 11 tests pass, including the below-min / above-max fuzz revert ranges and the in-range fuzz success range
- [x] Boundary success: `decimals = 6` and `decimals = 18`
- [x] Point reverts: `decimals = 5` and `decimals = 19` revert `InvalidDecimals(d)`
- [x] Range fuzz reverts: `[0, 5]` and `[19, 255]`
- [x] Range fuzz successes: `[6, 18]`
- [x] Storage round-trip: `vm.load(decimalsSlot())` equals param
- [x] Event integrity: `B20Created.decimals` equals input on `expectEmit` AND on recorded-log decode
- [x] Stablecoin regression: still `decimals() == 6`
- [x] Existing `test_createB20_revert_unsupportedVersion_security` updated to track the constant (now `2`)

### DIFF
--- a/src/lib/B20FactoryLib.sol
+++ b/src/lib/B20FactoryLib.sol
@@ -15,7 +15,9 @@ library B20FactoryLib {
     uint8 internal constant B20_STABLECOIN_CREATE_PARAMS_VERSION = 1;
 
     /// @notice Encoding version carried as the leading byte of a `B20SecurityCreateParams` blob.
-    uint8 internal constant B20_SECURITY_CREATE_PARAMS_VERSION = 1;
+    ///         Bumped from `1` to `2` when the struct gained the `decimals` field
+    ///         (previously hardcoded to `6` by the factory's security arm).
+    uint8 internal constant B20_SECURITY_CREATE_PARAMS_VERSION = 2;
 
     /// @notice Encoding version carried as the leading byte of a `B20StablecoinEventParams` blob.
     uint8 internal constant B20_STABLECOIN_EVENT_PARAMS_VERSION = 1;
@@ -102,14 +104,20 @@ library B20FactoryLib {
     /// @param name         ERC-20 token name.
     /// @param symbol       ERC-20 token symbol.
     /// @param initialAdmin Initial holder of `DEFAULT_ADMIN_ROLE`, or `address(0)` to deploy admin-less.
-    function encodeSecurityCreateParams(string memory name, string memory symbol, address initialAdmin)
+    /// @param decimals     ERC-20 `decimals` value. Must be in `[B20Constants.MIN_ASSET_DECIMALS, B20Constants.MAX_ASSET_DECIMALS]`;
+    ///                     out-of-range values cause the factory to revert with `InvalidDecimals`.
+    function encodeSecurityCreateParams(string memory name, string memory symbol, address initialAdmin, uint8 decimals)
         internal
         pure
         returns (bytes memory)
     {
         return abi.encode(
             IB20Factory.B20SecurityCreateParams({
-                version: B20_SECURITY_CREATE_PARAMS_VERSION, name: name, symbol: symbol, initialAdmin: initialAdmin
+                version: B20_SECURITY_CREATE_PARAMS_VERSION,
+                name: name,
+                symbol: symbol,
+                initialAdmin: initialAdmin,
+                decimals: decimals
             })
         );
     }
--- a/src/lib/B20FactoryLib.sol
+++ b/src/lib/B20FactoryLib.sol
@@ -15,9 +15,7 @@ library B20FactoryLib {
     uint8 internal constant B20_STABLECOIN_CREATE_PARAMS_VERSION = 1;
 
     /// @notice Encoding version carried as the leading byte of a `B20SecurityCreateParams` blob.
-    ///         Bumped from `1` to `2` when the struct gained the `decimals` field
-    ///         (previously hardcoded to `6` by the factory's security arm).
-    uint8 internal constant B20_SECURITY_CREATE_PARAMS_VERSION = 2;
+    uint8 internal constant B20_SECURITY_CREATE_PARAMS_VERSION = 1;
 
     /// @notice Encoding version carried as the leading byte of a `B20StablecoinEventParams` blob.
     uint8 internal constant B20_STABLECOIN_EVENT_PARAMS_VERSION = 1;
--- a/test/lib/mocks/MockB20.sol
+++ b/test/lib/mocks/MockB20.sol
@@ -161,14 +161,14 @@ abstract contract MockB20 is IB20 {
         return MockB20Storage.layout().symbol;
     }
 
-    /// @notice Default-variant decimals are fixed at 18. The security variant
-    ///         overrides this to read its per-token configured decimals from
-    ///         `MockB20SecurityStorage`, which requires `view` (not `pure`)
-    ///         mutability — so the base declares `view` to keep the override
-    ///         legal under Solidity's tightening rules.
-    function decimals() external view virtual returns (uint8) {
-        return 18;
-    }
+    /// @notice ERC-20 `decimals`. Unimplemented at the base; each variant
+    ///         must provide its own override. `MockB20Stablecoin.decimals()`
+    ///         returns `pure 6`; `MockB20Security.decimals()` returns the
+    ///         per-token value stored at `MockB20SecurityStorage.decimalsSlot()`
+    ///         (declared `view` for the storage read). Leaving the base
+    ///         abstract forces new variants to make an explicit decision and
+    ///         avoids accidentally inheriting a stale default.
+    function decimals() external view virtual returns (uint8);
 
     function totalSupply() external view returns (uint256) {
         return MockB20Storage.layout().totalSupply;
--- a/test/lib/mocks/MockB20.sol
+++ b/test/lib/mocks/MockB20.sol
@@ -161,8 +161,12 @@ abstract contract MockB20 is IB20 {
         return MockB20Storage.layout().symbol;
     }
 
-    /// @notice Default-variant decimals are fixed at 18.
-    function decimals() external pure virtual returns (uint8) {
+    /// @notice Default-variant decimals are fixed at 18. The security variant
+    ///         overrides this to read its per-token configured decimals from
+    ///         `MockB20SecurityStorage`, which requires `view` (not `pure`)
+    ///         mutability — so the base declares `view` to keep the override
+    ///         legal under Solidity's tightening rules.
+    function decimals() external view virtual returns (uint8) {
         return 18;
     }
 
--- a/test/lib/mocks/MockB20Factory.sol
+++ b/test/lib/mocks/MockB20Factory.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.20;
 
 import {Vm} from "forge-std/Vm.sol";
 
+import {B20Constants} from "src/lib/B20Constants.sol";
 import {IB20Factory} from "src/interfaces/IB20Factory.sol";
 import {B20FactoryLib} from "src/lib/B20FactoryLib.sol";
 import {StdPrecompiles} from "src/StdPrecompiles.sol";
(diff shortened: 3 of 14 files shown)


src/lib/B20FactoryLib.sol
```diff
@@ -15,7 +15,9 @@ library B20FactoryLib {
     uint8 internal constant B20_STABLECOIN_CREATE_PARAMS_VERSION = 1;
 
     /// @notice Encoding version carried as the leading byte of a `B20SecurityCreateParams` blob.
-    uint8 internal constant B20_SECURITY_CREATE_PARAMS_VERSION = 1;
+    ///         Bumped from `1` to `2` when the struct gained the `decimals` field
+    ///         (previously hardcoded to `6` by the factory's security arm).
+    uint8 internal constant B20_SECURITY_CREATE_PARAMS_VERSION = 2;
```
ilikesymmetry: don't version keep as `1` for now given we are prerelease

test/lib/mocks/MockB20.sol
```diff
@@ -161,8 +161,12 @@ abstract contract MockB20 is IB20 {
...
-    /// @notice Default-variant decimals are fixed at 18.
-    function decimals() external pure virtual returns (uint8) {
+    /// @notice Default-variant decimals are fixed at 18. The security variant
+    ///         overrides this to read its per-token configured decimals from
+    ///         `MockB20SecurityStorage`, which requires `view` (not `pure`)
+    ///         mutability — so the base declares `view` to keep the override
+    ///         legal under Solidity's tightening rules.
+    function decimals() external view virtual returns (uint8) {
         return 18;
     }
```
ilikesymmetry: thinking we should actually leave this unimplemented and force inheriting contracts to implement and override?

test/lib/mocks/MockB20Factory.sol
ilikesymmetry, line 1: also add new `uint8 decimals` to B20Created event and hardcode to `6` for stablecoin variant

test/lib/mocks/MockB20Storage.sol
```diff
@@ -353,6 +361,12 @@ library MockB20SecurityStorage {
         // Scaled by WAD_PRECISION (1e18). Stored value of 0 is
         // interpreted as WAD by the read surface.
         uint256 sharesToTokensRatio;
+        // ---------- Decimals ----------
+        // Per-token ERC-20 `decimals`. Written by the factory at
+        // creation from the `B20SecurityCreateParams.decimals` field
+        // (validated to [MIN_ASSET_DECIMALS, MAX_ASSET_DECIMALS]) and
+        // never mutated afterwards.
+        uint8 decimals;
```
ilikesymmetry, line 369: let's actually make `decimals` the first offset in the struct (above share ratio) and move everything down one versus inserting in the middle

test/unit/B20Factory/createB20_security_decimals.t.sol
```diff
@@ -0,0 +1,222 @@
...
+///
+/// @dev Asserts the boundary values, the out-of-range revert with the new
+///      `InvalidDecimals(uint8)` error, fuzz coverage over the full
+///      below-min / above-max / in-range ranges, the `B20Created` event's
+///      `decimals` field reflecting the input, the storage round-trip via
+///      the `MockB20SecurityStorage.decimalsSlot()` paired-slot pattern,
+///      and the stablecoin variant's regression-free hardcoded `6`.
+contract B20FactoryCreateB20SecurityDecimalsTest is B20FactoryTest {
+    /*//////////////////////////////////////////////////////////////
+                          BOUNDARY SUCCESSES
```
ilikesymmetry: just like with other tests, put the success cases after the revert cases. reference other tests for style

test/unit/B20Factory/createToken.t.sol
```diff
@@ -683,14 +686,23 @@ contract B20FactoryCreateB20Test is B20FactoryTest {
...
-    /// @dev Both variants return 6.
+    /// @notice Verifies stablecoin decimals are hardcoded at `6` and security decimals
+    ///         reflect the value passed at creation (rather than coming from address bytes).
+    /// @dev    Stablecoin decimals are immutable and not configurable. Security decimals
+    ///         are per-token (see `B20SecurityCreateParams.decimals`); the default
+    ///         `_securityParams()` helper passes `MIN_ASSET_DECIMALS` (`6`) so this test
+    ///         still pins the legacy-behavior case. Full range coverage lives in
+    ///         `createB20_security_decimals.t.sol`.
     function test_createB20_success_decimalsFixedByVariant(address caller, bytes32 salt) public {
         _assumeValidCaller(caller);
```
ilikesymmetry: this whole test needs to be changed now actually given decimals are not fixed by variant, only stablecoin. security/asset is dynamic

src/lib/B20FactoryLib.sol
ilikesymmetry: Addressed in 4ff6636 — reverted `B20_SECURITY_CREATE_PARAMS_VERSION` to `1` in `src/lib/B20FactoryLib.sol` and updated the matching "Currently `1`" natspec on `B20SecurityCreateParams.version` in `src/interfaces/IB20Factory.sol` plus the version-pin test in `test/unit/B20FactoryLib/encodeSecurityCreateParams.t.sol`.

test/lib/mocks/MockB20.sol
ilikesymmetry: Addressed in 4ff6636 — `MockB20.decimals()` is now abstract (declared `external view virtual returns (uint8);` with no body). Both variants already had concrete overrides: `MockB20Stablecoin.decimals()` returns `pure 6` and `MockB20Security.decimals()` returns the per-token slot. Bonus side effect: the prior interface-coverage orphan on `MockB20.decimals` is gone (verified locally via `scripts/check-coverage.py`).

test/lib/mocks/MockB20Factory.sol
ilikesymmetry: Already in place. The `B20Created` event was updated in this PR to include `uint8 decimals` (see `src/interfaces/IB20Factory.sol` line 113), and `MockB20Factory.createB20` hardcodes `decimals = 6` on the stablecoin branch (line 147) and threads it through the emit (line 192). Stablecoin decimals coverage is pinned in `test/unit/B20Factory/createB20_security_decimals.t.sol::test_createB20_success_stablecoin_decimalsStillHardcoded`. Let me know if you wanted something more — happy to add a dedicated event-pin test for the stablecoin path if useful.

test/lib/mocks/MockB20Storage.sol
ilikesymmetry: Addressed in 4ff6636 — `decimals` is now the first field of `MockB20SecurityStorage.Layout` (slot 0), with `sharesToTokensRatio` moved to slot 1 and the offset constants / natspec updated to match. Also reordered the assertions in `test/unit/storage/B20SecurityFullLayout.t.sol` and updated its slot-coverage comment block. `USED_ANNOUNCEMENT_IDS_OFFSET` and `IDENTIFIERS_OFFSET` stayed at 2 / 3 (mappings re-derive, not a layout break).

test/unit/B20Factory/createB20_security_decimals.t.sol
ilikesymmetry: Addressed in 4ff6636 — reordered `test/unit/B20Factory/createB20_security_decimals.t.sol` so the OUT-OF-RANGE REVERTS section now precedes BOUNDARY SUCCESSES / IN-RANGE FUZZ / STORAGE ROUND-TRIP / EVENT-VS-STATE COHERENCE / STABLECOIN REGRESSION SANITY, matching the revert-then-success layout used elsewhere in `test/unit/B20Factory/`.

test/unit/B20Factory/createToken.t.sol
ilikesymmetry: Addressed in 4ff6636 — deleted `test_createB20_success_decimalsFixedByVariant` from `test/unit/B20Factory/createToken.t.sol` entirely. The stablecoin-fixed-at-6 assertion now lives in `test_createB20_success_stablecoin_decimalsStillHardcoded` (in `createB20_security_decimals.t.sol`), and the security side is covered exhaustively by the in-range / boundary / fuzz tests in the same file — keeping a renamed two-line stub would just duplicate them. Also dropped the stale "bumped from 1 to 2" line from the `test_createB20_revert_unsupportedVersion_security` natspec.

rayyan224 (approved):

eric-ships (approved):

test/lib/mocks/MockB20Storage.sol
```diff
@@ -374,9 +391,13 @@ library MockB20SecurityStorage {
     // unrelated locations); the factory uses these as base slots when
     // deriving member slots via `keccak256(abi.encode(key, baseSlot))`.
 
-    uint256 internal constant SHARES_TO_TOKENS_RATIO_OFFSET = 0;
-    uint256 internal constant USED_ANNOUNCEMENT_IDS_OFFSET = 1;
-    uint256 internal constant IDENTIFIERS_OFFSET = 2;
+    // `decimals` (a `uint8`) sits alone in slot 0; the remaining 31
```
stevieraykatz: This comment doesn't make sense.

stevieraykatz: We're expecting to future-pack the slot? Seems weird to call out.

ilikesymmetry: agreed, will remove

ilikesymmetry: Addressed in b99bad3.

stevieraykatz (approved):

PR #133, opened 2026-06-02
fix(b20): consume allowance on privileged transferFrom (L-04)

ilikesymmetry:
## Summary

Audit finding **L-04** — *Inconsistent Allowance Deduction*. A privileged (factory bootstrap-window) `transferFrom` handled the spender allowance inconsistently between the Rust precompile and the Solidity reference.

The Rust precompile **consumes the spender's allowance on every `transferFrom`** — there is no `privileged` exception for allowance accounting. A privileged caller only bypasses the **executor-policy** check; an infinite (`type(uint256).max`) allowance is never decremented. The Solidity reference (`MockB20`) instead skipped the **entire** allowance block during the bootstrap window — neither checking nor decrementing — so a privileged caller could move tokens with no allowance at all.

This aligns the Solidity reference to the Rust behavior: **consume allowance unconditionally**, gating only the executor-policy check on the bootstrap window.

## Changes

- `transferFrom` / `transferFromWithMemo`: call `_consumeAllowance` unconditionally (check + decrement, infinite-safe); keep the executor-policy check behind `!_isPrivileged() && msg.sender != from`.
- Corrected the `_transfer` natspec to state that allowance consumption is unconditional and only the executor-policy check honors the bootstrap bypass.

## Tests

Added fuzz regression tests in `transferFrom.t.sol` and `transferFromWithMemo.t.sol`:
- `*_success_privileged_decrementsAllowance` — a privileged call decrements the allowance by the spent amount, exactly like an ordinary call (paired storage-slot assertion). Fails before this change (privileged left allowance untouched), passes after.
- `*_revert_privileged_insufficientAllowance` — a privileged call still reverts `InsufficientAllowance` when the allowance is insufficient.

## Test plan

- [x] `forge fmt --check` clean
- [x] `forge test` — 606 passed, 0 failed, 0 skipped
- [ ] Confirm parity with the Rust precompile before merge (no `privileged` allowance exception there)

> Draft for review of the chosen direction: align Solidity to the existing Rust behavior (always consume) rather than changing Rust to skip.

### DIFF
--- a/test/lib/mocks/MockB20.sol
+++ b/test/lib/mocks/MockB20.sol
@@ -186,20 +186,21 @@ abstract contract MockB20 is IB20 {
         returns (bool)
     {
         _requireNonZeroActors(from, to);
-        if (!_isPrivileged()) {
-            // Allowance is consumed unconditionally outside the factory
-            // bootstrap window. Matches OZ ERC20 and the Rust precompile,
-            // both of which carve no exception for `msg.sender == from`.
-            _consumeAllowance(from, msg.sender, amount);
-            if (msg.sender != from) {
-                // Read the executor policy ID out of the transfer-side packed
-                // slot. Cold here; warm by the time _transfer reads the same
-                // slot for sender + receiver. Skipped when the caller is the
-                // owner — sender-policy already covers `from` inside _transfer.
-                uint64 executorPolicyId = MockB20Storage.layout().transferPolicyIds.executor;
-                if (!IPolicyRegistry(POLICY_REGISTRY).isAuthorized(executorPolicyId, msg.sender)) {
-                    revert PolicyForbids(TRANSFER_EXECUTOR_POLICY, executorPolicyId);
-                }
+        // Allowance is consumed unconditionally — including during the factory
+        // bootstrap window (`_isPrivileged()`). Matches the Rust precompile,
+        // which carves no `privileged` exception for allowance accounting
+        // (BOP-230 / L-04); only the executor-policy check below is bypassed
+        // for a privileged caller. An infinite allowance is still not
+        // decremented (handled inside `_consumeAllowance`).
+        _consumeAllowance(from, msg.sender, amount);
+        if (!_isPrivileged() && msg.sender != from) {
+            // Read the executor policy ID out of the transfer-side packed
+            // slot. Cold here; warm by the time _transfer reads the same
+            // slot for sender + receiver. Skipped when the caller is the
+            // owner — sender-policy already covers `from` inside _transfer.
+            uint64 executorPolicyId = MockB20Storage.layout().transferPolicyIds.executor;
+            if (!IPolicyRegistry(POLICY_REGISTRY).isAuthorized(executorPolicyId, msg.sender)) {
+                revert PolicyForbids(TRANSFER_EXECUTOR_POLICY, executorPolicyId);
             }
         }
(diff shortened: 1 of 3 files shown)


test/lib/mocks/MockB20.sol
```diff
@@ -235,7 +241,13 @@ abstract contract MockB20 is IB20 {
         returns (bool)
     {
         _requireNonZeroActors(from, to);
-        if (!_isPrivileged()) {
+        if (_isPrivileged()) {
+            // Factory bootstrap window: the allowance check is still enforced — a
+            // privileged caller cannot move tokens it lacks an allowance for
+            // (BOP-230 / L-04) — but the decrement and the executor-policy check
+            // are bypassed, mirroring the Rust precompile.
+            _checkAllowance(from, msg.sender, amount);
```
roger-bai-coinbase: This is still incorrect with the Rust code. Allowances are consumed for privileged calls there

test/lib/mocks/MockB20.sol
```diff
@@ -702,10 +705,12 @@ abstract contract MockB20 is IB20 {
     ///      every external caller (`transfer`, `transferFrom`,
     ///      `transferWithMemo`, `transferFromWithMemo`) before reaching
     ///      this helper. `transferFrom` / `transferFromWithMemo`
-    ///      additionally consume allowance and check the executor
-    ///      policy in their bodies before calling here; both of those
-    ///      checks ALSO honor the bootstrap bypass, consistent with
-    ///      the policy bypass below.
+    ///      additionally consume the allowance (unconditionally —
+    ///      including in the bootstrap window, matching the Rust
+    ///      precompile, see BOP-230 / L-04) and check the executor
```
ilikesymmetry, line 710: don't add references to tickets, PRs, or finding ids

roger-bai-coinbase (approved):

eric-ships (approved):
## Review

Reviewed the two open items from the draft checklist.

### (1) Rust parity confirmed

`transfer_from` in `crates/common/precompiles/src/common/ops/transferable.rs`:

```rust
let allowance = self.accounting().allowance(from, spender)?;
let is_infinite = allowance == U256::MAX;
if !is_infinite && allowance < amount { return Err(InsufficientAllowance) }
if !privileged && spender != from { /* executor policy */ }
self.transfer_inner(from, to, amount, privileged)?;
if is_infinite { return Ok(()); }
self.accounting_mut().set_allowance(from, spender, allowance - amount)
```

Allowance is read, checked, and decremented unconditionally. `privileged` only gates the executor-policy branch. There is a `transfer_from_privileged_skips_executor_policy` unit test in the Rust tree pinning exactly this. The PR's fix matches the Rust behavior precisely — the unchecked box can be ticked.

### (2) `_consumeAllowance` when `from == msg.sender` is safe

`_consumeAllowance(owner, spender, amount)` reads `allowances[owner][spender]`. When `from == msg.sender`, this becomes `allowances[from][from]`, which defaults to `0`. So `transferFrom(self, to, amount)` with `amount > 0` and no self-approval reverts `InsufficientAllowance` — identical to the Rust path, which applies the same allowance read without any `spender != from` guard. The behavior is consistent and intentional.

### Overall

The logic change is minimal and surgical. The two fuzz tests directly pin the before/after behavior (both check + decrement, and insufficient-allowance revert). The natspec on `_transfer` is accurate. LGTM.

PR #140, opened 2026-06-03
fix(BOP-274): port Rust parity fixes into Solidity mocks 

stevieraykatz:
## Summary

Brings two Solidity-mock behaviors back into parity with the Rust precompile after Roger flagged the divergences:

- **Mint receiver policy** — `_mint` no longer skips `MINT_RECEIVER_POLICY` when called from the factory bootstrap window. Rust enforces the policy unconditionally; the policy itself defaults to `ALWAYS_ALLOW_ID` for an unconfigured slot, so existing bootstrap flows are unaffected.
- **Burn-blocked target check** — `burnBlocked` no longer skips the "target is blocked" check during bootstrap. Same shape as the mint fix: unconditional enforcement, default-allow when unconfigured.

The third subtask (`is_announcement_active` view) was explored but dropped — `announce(...)` only self-delegatecalls, so there is no external reentrancy surface that would want to detect a bracket-in-progress. 

Linear: BOP-274.

## Test plan

- [x] `forge build` clean
- [x] `forge test` — 607 tests pass, no regressions


test/lib/mocks/MockB20Asset.sol
```diff
@@ -64,6 +64,28 @@ contract MockB20Asset is MockB20, IB20Asset {
...
+    ///
+    ///      Declared as a contract-level state variable (not an
+    ///      ERC-7201 namespaced struct field) because transient storage
+    ///      lives in its own opcode-distinct address space — slot
+    ///      indices here can't collide with the regular-storage layout
+    ///      written by the factory bootstrap. The Rust precompile
+    ///      exposes the same value via a runtime context flag rather
+    ///      than a storage slot, so there is no persistent layout the
+    ///      Rust impl needs to mirror.
+    bool internal transient _announcementActive;
```
stevieraykatz: the transient storage location is only available in solidity as of 0.8.28 but our pragma is open to 0.8.20 (not good). we could:
1. bump the required version for the whole repo to 0.8.28, or 
2. use an assembly tstore pattern (supported as of 0.8.24)
3. use a regular storage param

I'm partial to bumping the pragma to at least 0.8.24 and using tstorage (more closely reflects the rust revm memory impl), but I could also see us not bringing this feature to parity.

amiecorso: I think bumping the pragma to support the most expressive/accurate mock implementation makes sense to me? If there were genuine compatibility concerns for other devs importing/working with this source code I'd hesitate more but given that these aren't production implementations I don't think it matters?

src/interfaces/IB20Asset.sol
```diff
@@ -109,6 +109,20 @@ interface IB20Asset is IB20 {
...
 
+    /// @notice Whether an announcement bracket is currently open on this token. True from the
+    ///         start of `announce` through the dispatch of every entry in `internalCalls` and
+    ///         until `EndAnnouncement` fires; false at all other times, including before the
+    ///         first `announce` call and after the bracket closes. Resets per transaction:
+    ///         a revert anywhere in the bracket leaves no observable side effect on this view.
+    ///
+    /// @dev    Intended for inner-call contracts dispatched via `internalCalls` (or other
+    ///         contracts they reach) to detect they are executing inside an announcement
+    ///         bracket — useful for issuance, multiplier, and metadata flows whose policy
```
ilikesymmetry: I'm not sold on this

ilikesymmetry: Even if we wanted the state, it's not clear any external consumer would ever be able to use this given internal calls are internal so no external entity gets control flow when announcement is active

ilikesymmetry: But I'm also not sold the state is even useful given no code is actually using it seemingly?

PR #155, opened 2026-06-10
docs(smoke): README for the b20 precompile smoketest

amiecorso:
Adds a `script/smoke/README.md` for the b20 precompile smoketest. As a
first-time user I couldn't tell how to run the suite, what it needs, or how to
read its output, so this documents it end to end.

The suite is a runbook check that sends real transactions to a live JSON-RPC
endpoint where the b20 features are activated. It is not coupled to any
particular node: a remote Base fork (>= Beryl) or a node you run yourself (for
example a local build of `base/base`) both work, as long as the precompiles are
deployed and the features are switched on in the ActivationRegistry. You bring
the endpoint and two funded keys; the suite does not stand a node up or fund
anyone for you.

The README covers:

- what you need (an RPC endpoint with b20 active + two funded keys);
- running it (`make smoke-setup`, `.env`, the `make smoke-*` targets, the raw CLI);
- the five journeys and what each exercises;
- interpreting pass / fail / skip output;
- env / config knobs and troubleshooting.

Docs only, no code changes.


script/smoke/run-smoke-local.sh
```diff
@@ -0,0 +1,153 @@
...
+command -v curl  >/dev/null 2>&1 || die "curl not found"
+lsof -i ":$PORT" >/dev/null 2>&1 && die "port $PORT already in use. Set PORT=<other> or kill the listener."
+
+# ── Ensure the python venv exists ───────────────────────────────────────────────
+if [[ ! -x "$VENV/bin/python" ]]; then
+    log "creating smoke venv (one-time)…"
+    "$PYTHON" -m venv "$VENV"
+    "$VENV/bin/python" -m pip install --quiet --upgrade pip
+    "$VENV/bin/python" -m pip install --quiet -r "$SMOKE_DIR/requirements.txt"
+fi
```
stevieraykatz: this is done in smoke-setup

stevieraykatz (approved):

PR #160, opened 2026-06-10
feat: migrate smoke testing scripts to python

stevieraykatz:
The bash script was getting a bit unwieldy. This PR refactors the fork testing setup to use python. This is preferable now that we're requiring python for the smoke tests anyway. Until we have a formal `baseup + foundry` stack, this should be more maintainable.


script/fork/__main__.py
```diff
@@ -0,0 +1,264 @@
...
+
+# Canonical (name, id) order, mirroring the FEATURE_IDS array the bash runner activated in turn. Names are
+# the SKIP_ACTIVATE / log labels; ids gate each precompile via ActivationRegistry.activate(bytes32).
+FEATURES: list[tuple[str, bytes]] = [
+    ("B20_ASSET", bytes(FEATURE_B20_ASSET)),
+    ("POLICY_REGISTRY", bytes(FEATURE_POLICY_REGISTRY)),
+    ("B20_STABLECOIN", bytes(FEATURE_B20_STABLECOIN)),
+]
+
+ACTIVATE_SELECTOR = bytes(Web3.keccak(text="activate(bytes32)")[:4])
```
stevieraykatz, line 42: can we pull this from the abi instead of recalculating?

amiecorso (approved):
I asked my agent to run all of this locally and capture any problems or snags it ran into. Everything went smoothly and confirmed the test behavior is the same between the shell script and python. So looks great to me!

PR #174, opened 2026-07-22
feat: composite policies (UNION/INTERSECT) — interface, mock, and tests

rayyan224:
## Summary

Adds composite-policy support to the policy registry: the `IPolicyRegistry` interface surface, the `MockPolicyRegistry` reference implementation, and the full unit-test suite. (Includes the mock implementation merged via #175.)

A composite policy combines existing **simple** policies (ALLOWLIST/BLOCKLIST) under a logic gate. The gate rides the policy ID's top byte, so composites need no separate type storage.

### Interface (`IPolicyRegistry`)

- **`PolicyType`** gains `UNION` (2) and `INTERSECT` (3). Append-only: leaf discriminants (`BLOCKLIST=0`, `ALLOWLIST=1`) are unchanged.
- **`createCompositePolicy(address admin, PolicyType policyType, uint64[] childPolicyIds)`** — creates a UNION/INTERSECT over existing simple child policies. Children must exist and be simple (never a composite or a built-in sentinel); the child count must be in `[2, 4]`.
- **`updateComposite(uint64 policyId, uint64[] childPolicyIds)`** — full replace of the child set, re-validated exactly as at creation. The gate is fixed in the ID and cannot change.
- **`isAuthorized`** — composites evaluate **live** over current child membership: UNION is OR (any child authorizes; empty → false), INTERSECT is AND (every child authorizes; empty → true).
- **Event:** `CompositePolicyUpdated(policyId, updater, childPolicyIds)` — emitted on creation and every update, carrying the full post-update child set for single-log indexing. (Creation also emits the canonical `PolicyCreated` + `PolicyAdminUpdated`.)
- **Errors:** `ChildPoliciesOutsideOfRange(min, max)` when the child count is outside `[2, 4]`; `InvalidChildPolicy(childPolicyId)` when a child is a built-in sentinel or a composite. Reuses `PolicyNotFound` / `IncompatiblePolicyType` / `Unauthorized` / `ZeroAddress`.

### Mock implementation

- `MockPolicyRegistryStorage`: adds a `children` mapping (`uint64 => uint64[]`) at layout offset 4 (offsets 0–3 unchanged) plus a `childrenSlot` helper for Rust slot-math parity.
- `MockPolicyRegistry`: implements `createCompositePolicy` / `updateComposite` with the canonical revert precedence, and splits composite evaluation into `_isAuthorizedUnion` / `_isAuthorizedIntersect`. Child validation (`_requireCreatedSimplePolicies`) is two-pass so `PolicyNotFound` precedes `InvalidChildPolicy`.

### Tests

`test/unit/PolicyRegistry/`: composite create/update (reverts + success), sequential revert-order walks, `isAuthorized` UNION/INTERSECT/mixed-child/live-membership semantics, and built-in-sentinel-child rejection.

## Testing

`forge test` → 658 passed, 0 failed, 4 skipped (the 4 skipped are mock-only counter-overflow tests guarded by `vm.skip(livePrecompiles)`). `forge fmt --check` clean.

Generated with Claude Code

### DIFF
--- a/src/interfaces/IPolicyRegistry.sol
+++ b/src/interfaces/IPolicyRegistry.sol
@@ -58,9 +58,7 @@ interface IPolicyRegistry {
 
     /// @notice A composite policy was created or updated with fewer than the minimum number of
     ///         child policies. A composite must reference at least two simple policies.
-    /// @param provided Number of child policy IDs supplied.
-    /// @param minimum  Minimum required (2).
-    error TooFewChildPolicies(uint256 provided, uint256 minimum);
+    error TooFewChildPolicies();
 
     /// @notice Composite policies are not simple policies. Child policies must be existing
     ///         ALLOWLIST or BLOCKLIST policies;
@@ -99,7 +97,6 @@ interface IPolicyRegistry {
     ///
     /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`.
     /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is a composite gate.
-    /// @dev Panics with arithmetic overflow (Panic 0x11) when the policy counter has reached its maximum value.
     ///
     /// @param admin      Initial admin authorized to modify membership and transfer or renounce administration.
     /// @param policyType BLOCKLIST or ALLOWLIST.
--- a/src/interfaces/IPolicyRegistry.sol
+++ b/src/interfaces/IPolicyRegistry.sol
@@ -87,13 +87,8 @@ interface IPolicyRegistry {
     /// @notice One or more accounts had their BLOCKLIST membership set to `blocked` in a single batch.
     event BlocklistUpdated(uint64 indexed policyId, address indexed updater, bool blocked, address[] accounts);
 
-    /// @notice A composite policy (UNION or INTERSECT) was created over `childPolicyIds`.
-    event CompositePolicyCreated(
-        uint64 indexed policyId, address indexed creator, PolicyType policyType, uint64[] childPolicyIds
-    );
-
-    /// @notice A composite policy's child set was replaced in full with `childPolicyIds`. The event
-    ///         carries the complete post-update set.
+    /// @notice A composite policy's child set was set or replaced in full with `childPolicyIds`. Emitted
+    ///         on composite creation and on every subsequent update; carries the complete post-update set.
     event CompositePolicyUpdated(uint64 indexed policyId, address indexed updater, uint64[] childPolicyIds);
 
     /*//////////////////////////////////////////////////////////////
--- a/src/interfaces/IPolicyRegistry.sol
+++ b/src/interfaces/IPolicyRegistry.sol
@@ -111,8 +111,7 @@ interface IPolicyRegistry {
     /// @notice Creates a new simple policy seeded with `accounts` as initial members. Permissionless.
     ///
     /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`. Takes precedence over `BatchSizeTooLarge`.
-    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is a composite gate (UNION or
-    ///      INTERSECT); use `createCompositePolicy` for those. Takes precedence over `BatchSizeTooLarge`.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is a composite policyType.
     /// @dev Reverts with `BatchSizeTooLarge` when `accounts.length` exceeds the registry limit.
     /// @dev Panics with arithmetic overflow (Panic 0x11) when the policy counter has reached its maximum value.
     ///
--- a/src/interfaces/IPolicyRegistry.sol
+++ b/src/interfaces/IPolicyRegistry.sol
@@ -10,13 +10,19 @@ interface IPolicyRegistry {
                                   TYPES
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice Policy type discriminator.
+    /// @notice Policy type discriminator. Discriminants are append-only: BLOCKLIST and ALLOWLIST
+    ///         keep values 0 and 1, so the composite gates ride the policy ID top byte without
+    ///         shifting the existing leaf types.
     ///
-    /// @param BLOCKLIST Authorized unless in the policy's set.
-    /// @param ALLOWLIST Authorized only if in the policy's set.
+    /// @param BLOCKLIST Leaf. Authorized unless in the policy's set.
+    /// @param ALLOWLIST Leaf. Authorized only if in the policy's set.
+    /// @param UNION     Composite. Authorized if authorized under ANY operand (logical OR).
+    /// @param INTERSECT Composite. Authorized only if authorized under EVERY operand (logical AND).
     enum PolicyType {
         BLOCKLIST,
-        ALLOWLIST
+        ALLOWLIST,
+        UNION,
+        INTERSECT
     }
 
     /*//////////////////////////////////////////////////////////////
@@ -48,6 +54,16 @@ interface IPolicyRegistry {
     /// @notice `finalizeUpdateAdmin` was called with no pending admin staged.
     error NoPendingAdmin();
 
+    /// @notice A composite policy was created or updated with an empty operand set. A composite must
+    ///         reference at least one operand, so the never-revert fold evaluator need not define
+    ///         empty-UNION / empty-INTERSECT semantics.
+    error EmptyOperandSet();
+
+    /// @notice A composite operand is not a flat leaf. Operands must be existing ALLOWLIST or
+    ///         BLOCKLIST policies; composites of composites are rejected (flat-only).
+    /// @param operandId The offending operand policy ID.
+    error InvalidOperand(uint64 operandId);
+
     /*//////////////////////////////////////////////////////////////
                                  EVENTS
     //////////////////////////////////////////////////////////////*/
@@ -68,13 +84,25 @@ interface IPolicyRegistry {
     /// @notice One or more accounts had their BLOCKLIST membership set to `blocked` in a single batch.
     event BlocklistUpdated(uint64 indexed policyId, address indexed updater, bool blocked, address[] accounts);
 
+    /// @notice A composite policy (UNION or INTERSECT) was created over `operands`.
+    event CompositePolicyCreated(
+        uint64 indexed policyId, address indexed creator, PolicyType policyType, uint64[] operands
+    );
+
+    /// @notice A composite policy's operand set was replaced in full with `operands`. The event
+    ///         carries the complete post-update set, so an indexer reconstructs current state from a
+    ///         single log with no delta folding.
+    event CompositeOperandsUpdated(uint64 indexed policyId, address indexed updater, uint64[] operands);
+
     /*//////////////////////////////////////////////////////////////
                             POLICY CREATION
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice Creates a new policy with no initial members. Permissionless.
+    /// @notice Creates a new leaf policy with no initial members. Permissionless.
     ///
     /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is a composite gate (UNION or
+    ///      INTERSECT); use `createCompositePolicy` for those.
     /// @dev Panics with arithmetic overflow (Panic 0x11) when the policy counter has reached its maximum value.
     ///
     /// @param admin      Initial admin authorized to modify membership and transfer or renounce administration.
@@ -83,9 +111,11 @@ interface IPolicyRegistry {
     /// @return newPolicyId The newly assigned policy ID.
     function createPolicy(address admin, PolicyType policyType) external returns (uint64 newPolicyId);
 
-    /// @notice Creates a new policy seeded with `accounts` as initial members. Permissionless.
+    /// @notice Creates a new leaf policy seeded with `accounts` as initial members. Permissionless.
     ///
     /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`. Takes precedence over `BatchSizeTooLarge`.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is a composite gate (UNION or
+    ///      INTERSECT); use `createCompositePolicy` for those. Takes precedence over `BatchSizeTooLarge`.
     /// @dev Reverts with `BatchSizeTooLarge` when `accounts.length` exceeds the registry limit.
     /// @dev Panics with arithmetic overflow (Panic 0x11) when the policy counter has reached its maximum value.
     ///
@@ -95,6 +125,30 @@ interface IPolicyRegistry {
     ///
     /// @return newPolicyId The newly assigned policy ID.
     function createPolicyWithAccounts(address admin, PolicyType policyType, address[] calldata accounts)
+        external
+        returns (uint64 newPolicyId);
+
+    /// @notice Creates a new composite policy that combines existing leaf policies under a logic gate.
+    ///         Permissionless. A UNION authorizes an account if ANY operand authorizes it; an INTERSECT
+    ///         authorizes only if EVERY operand does. `isAuthorized` folds the operands and never reverts.
+    ///
+    /// @dev Operands are flat-only: each must be an existing ALLOWLIST or BLOCKLIST policy, never another
+    ///      composite. Because operands reference smaller, already-assigned IDs, the reference graph is a
+    ///      DAG and cannot cycle. The operand set is capped at the composite operand limit.
+    /// @dev Reverts with `IncompatiblePolicyType` when `gate` is not UNION or INTERSECT.
+    /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`.
+    /// @dev Reverts with `EmptyOperandSet` when `operands` is empty.
+    /// @dev Reverts with `BatchSizeTooLarge` when `operands.length` exceeds the composite operand limit.
+    /// @dev Reverts with `PolicyNotFound` when any operand does not exist.
+    /// @dev Reverts with `InvalidOperand` when any operand is itself a composite (not a flat leaf).
+    /// @dev Panics with arithmetic overflow (Panic 0x11) when the policy counter has reached its maximum value.
+    ///
+    /// @param admin    Initial admin authorized to update operands and transfer or renounce administration.
+    /// @param gate     UNION or INTERSECT.
+    /// @param operands Existing leaf policy IDs to combine.
+    ///
+    /// @return newPolicyId The newly assigned composite policy ID.
+    function createCompositePolicy(address admin, PolicyType gate, uint64[] calldata operands)
         external
         returns (uint64 newPolicyId);
 
@@ -153,6 +207,30 @@ interface IPolicyRegistry {
     /// @param blocked  Membership state to apply to every account in the batch.
     /// @param accounts Accounts to update.
     function updateBlocklist(uint64 policyId, bool blocked, address[] calldata accounts) external;
+
+    /// @notice Replaces a composite policy's operand set in full with `operands`. This is a whole-set
+    ///         overwrite, not an incremental edit: the caller sends the complete desired operand list
+    ///         and it is re-validated exactly as at creation. The gate (UNION or INTERSECT) is fixed in
+    ///         the policy ID and cannot change; a different gate requires a new composite.
+    ///
+    /// @dev Overwrite semantics are last-write-wins. The function carries no expected-version guard, so
+    ///      concurrent full-set submissions by shared admins silently clobber one another; serialize
+    ///      edits through the governing multisig or approval flow. Read the current operands before
+    ///      editing.
+    /// @dev Guard order: `PolicyNotFound` -> `IncompatiblePolicyType` -> `Unauthorized` ->
+    ///      `EmptyOperandSet` / `BatchSizeTooLarge` -> per-operand `PolicyNotFound` / `InvalidOperand`.
+    /// @dev Reverts with `PolicyNotFound` when `policyId` does not exist.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyId` is not a composite (UNION or INTERSECT).
+    /// @dev Reverts with `Unauthorized` when the caller is not the current admin. A renounced composite
+    ///      (admin `address(0)`) can never be updated.
+    /// @dev Reverts with `EmptyOperandSet` when `operands` is empty; there is no clear-the-list path.
+    /// @dev Reverts with `BatchSizeTooLarge` when `operands.length` exceeds the composite operand limit.
+    /// @dev Reverts with `PolicyNotFound` when any operand does not exist.
+    /// @dev Reverts with `InvalidOperand` when any operand is itself a composite (not a flat leaf).
+    ///
+    /// @param policyId Composite policy to update.
+    /// @param operands Complete new operand set of existing leaf policy IDs.
+    function updateCompositeOperands(uint64 policyId, uint64[] calldata operands) external;
 
     /*//////////////////////////////////////////////////////////////
                          AUTHORIZATION QUERIES
--- a/src/interfaces/IPolicyRegistry.sol
+++ b/src/interfaces/IPolicyRegistry.sol
@@ -56,9 +56,12 @@ interface IPolicyRegistry {
     /// @notice `finalizeUpdateAdmin` was called with no pending admin staged.
     error NoPendingAdmin();
 
-    /// @notice A composite policy was created or updated with fewer than the minimum number of
-    ///         child policies. A composite must reference at least two simple policies.
-    error TooFewChildPolicies();
+    /// @notice A composite policy was created or updated with a child-policy count outside the
+    ///         permitted range. A composite must reference between `min` and `max` simple policies,
+    ///         inclusive.
+    /// @param min Minimum number of child policies permitted.
+    /// @param max Maximum number of child policies permitted.
+    error ChildPoliciesOutsideOfRange(uint256 min, uint256 max);
 
     /// @notice Composite policies are not simple policies. Child policies must be existing
     ///         ALLOWLIST or BLOCKLIST policies;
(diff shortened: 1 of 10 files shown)


src/interfaces/IPolicyRegistry.sol
```diff
@@ -83,9 +111,11 @@ interface IPolicyRegistry {
     /// @return newPolicyId The newly assigned policy ID.
     function createPolicy(address admin, PolicyType policyType) external returns (uint64 newPolicyId);
 
-    /// @notice Creates a new policy seeded with `accounts` as initial members. Permissionless.
+    /// @notice Creates a new leaf policy seeded with `accounts` as initial members. Permissionless.
```
rayyan224: Remove leaf here

src/interfaces/IPolicyRegistry.sol
```diff
@@ -68,13 +84,25 @@ interface IPolicyRegistry {
     /// @notice One or more accounts had their BLOCKLIST membership set to `blocked` in a single batch.
     event BlocklistUpdated(uint64 indexed policyId, address indexed updater, bool blocked, address[] accounts);
 
+    /// @notice A composite policy (UNION or INTERSECT) was created over `operands`.
+    event CompositePolicyCreated(
+        uint64 indexed policyId, address indexed creator, PolicyType policyType, uint64[] operands
```
ilikesymmetry: feel like I want to see the clarification of "policy" in there
```suggestion
        uint64 indexed policyId, address indexed creator, PolicyType policyType, uint64[] operandPolicies
```

src/interfaces/IPolicyRegistry.sol
```diff
@@ -98,6 +128,30 @@ interface IPolicyRegistry {
...
+    /// @dev Reverts with `PolicyNotFound` when any operand does not exist.
+    /// @dev Reverts with `InvalidOperand` when any operand is itself a composite (not a flat leaf).
+    /// @dev Panics with arithmetic overflow (Panic 0x11) when the policy counter has reached its maximum value.
+    ///
+    /// @param admin    Initial admin authorized to update operands and transfer or renounce administration.
+    /// @param gate     UNION or INTERSECT.
+    /// @param operands Existing leaf policy IDs to combine.
+    ///
+    /// @return newPolicyId The newly assigned composite policy ID.
+    function createCompositePolicy(address admin, PolicyType gate, uint64[] calldata operands)
```
ilikesymmetry: prefer less new words:

```suggestion
    function createCompositePolicy(address admin, PolicyType policyType, uint64[] calldata operandPolicies)
```

src/interfaces/IPolicyRegistry.sol
```diff
@@ -154,6 +208,30 @@ interface IPolicyRegistry {
...
+    /// @dev Reverts with `Unauthorized` when the caller is not the current admin. A renounced composite
+    ///      (admin `address(0)`) can never be updated.
+    /// @dev Reverts with `EmptyOperandSet` when `operands` is empty; there is no clear-the-list path.
+    /// @dev Reverts with `BatchSizeTooLarge` when `operands.length` exceeds the composite operand limit.
+    /// @dev Reverts with `PolicyNotFound` when any operand does not exist.
+    /// @dev Reverts with `InvalidOperand` when any operand is itself a composite (not a flat leaf).
+    ///
+    /// @param policyId Composite policy to update.
+    /// @param operands Complete new operand set of existing leaf policy IDs.
+    function updateCompositeOperands(uint64 policyId, uint64[] calldata operands) external;
```
ilikesymmetry: "composite operands" doesn't pass my gut check for being easily intuitable on surface. What other options for "operand" have we considered? My previous comments of expanding to "operand policy" may just be a bandaid for a name that needs reworking, although I'm okay still keeping it. In general, I don't think we should ever use the word "operand" without immediately following it with "policy" for clarity.

src/interfaces/IPolicyRegistry.sol
ilikesymmetry: Do we even need this event? Why not just use `PolicyCreated` as-is and then combine with the event below for declaring operand policies updated?

src/interfaces/IPolicyRegistry.sol
```diff
@@ -10,13 +10,19 @@ interface IPolicyRegistry {
...
+    /// @notice Policy type discriminator. Discriminants are append-only: BLOCKLIST and ALLOWLIST
+    ///         keep values 0 and 1, so the composite gates ride the policy ID top byte without
+    ///         shifting the existing leaf types.
     ///
-    /// @param BLOCKLIST Authorized unless in the policy's set.
-    /// @param ALLOWLIST Authorized only if in the policy's set.
+    /// @param BLOCKLIST Leaf. Authorized unless in the policy's set.
+    /// @param ALLOWLIST Leaf. Authorized only if in the policy's set.
+    /// @param UNION     Composite. Authorized if authorized under ANY operand (logical OR).
+    /// @param INTERSECT Composite. Authorized only if authorized under EVERY operand (logical AND).
```
ilikesymmetry: I feel like saying "Composite." as the first explanation of the thing doesn't communicate value upfront. I think a clearer reader experience is to explain the thing in plain english first and sprinkle the vocabulary we need as we go or affirm at the end.

```suggestion
    /// @param UNION     Authorized if ANY operand policy is authorized (logical OR).
    /// @param INTERSECT Authorized only if EVERY operand policy is authorized (logical AND).
```

src/interfaces/IPolicyRegistry.sol
```diff
@@ -10,13 +10,19 @@ interface IPolicyRegistry {
                                   TYPES
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice Policy type discriminator.
+    /// @notice Policy type discriminator. Discriminants are append-only: BLOCKLIST and ALLOWLIST
+    ///         keep values 0 and 1, so the composite gates ride the policy ID top byte without
+    ///         shifting the existing leaf types.
```
stevieraykatz: a lil sloppy copy here. probably fine as: 

```suggestion
    /// @notice Policy type discriminator. Discriminants are append-only.
```

src/interfaces/IPolicyRegistry.sol
```diff
@@ -48,6 +54,16 @@ interface IPolicyRegistry {
     /// @notice `finalizeUpdateAdmin` was called with no pending admin staged.
     error NoPendingAdmin();
 
+    /// @notice A composite policy was created or updated with an empty operand set. A composite must
+    ///         reference at least one operand, so the never-revert fold evaluator need not define
```
stevieraykatz: minimum two? since we're not offering a NOT gate, we need two at a minimum to be meaningful.

src/interfaces/IPolicyRegistry.sol
```diff
@@ -98,6 +128,30 @@ interface IPolicyRegistry {
         external
         returns (uint64 newPolicyId);
 
+    /// @notice Creates a new composite policy that combines existing leaf policies under a logic gate.
+    ///         Permissionless. A UNION authorizes an account if ANY operand authorizes it; an INTERSECT
+    ///         authorizes only if EVERY operand does. `isAuthorized` folds the operands and never reverts.
```
stevieraykatz: remove fluff or move to @dev block. @notice should be one line generally.

src/interfaces/IPolicyRegistry.sol
```diff
@@ -48,6 +54,16 @@ interface IPolicyRegistry {
     /// @notice `finalizeUpdateAdmin` was called with no pending admin staged.
     error NoPendingAdmin();
 
+    /// @notice A composite policy was created or updated with an empty operand set. A composite must
+    ///         reference at least one operand, so the never-revert fold evaluator need not define
+    ///         empty-UNION / empty-INTERSECT semantics.
+    error EmptyOperandSet();
+
+    /// @notice A composite operand is not a flat leaf. Operands must be existing ALLOWLIST or
+    ///         BLOCKLIST policies; a composite may not reference another composite (flat-only).
```
stevieraykatz: Slim down.
```suggestion
    /// @notice Operands must be existing ALLOWLIST or BLOCKLIST policies.
```

src/interfaces/IPolicyRegistry.sol
```diff
@@ -68,13 +84,25 @@ interface IPolicyRegistry {
...
     event BlocklistUpdated(uint64 indexed policyId, address indexed updater, bool blocked, address[] accounts);
 
+    /// @notice A composite policy (UNION or INTERSECT) was created over `operands`.
+    event CompositePolicyCreated(
+        uint64 indexed policyId, address indexed creator, PolicyType policyType, uint64[] operands
+    );
+
+    /// @notice A composite policy's operand set was replaced in full with `operands`. The event
+    ///         carries the complete post-update set, so an indexer reconstructs current state from a
+    ///         single log with no delta folding.
```
stevieraykatz: the event carrying the full post-update set is nice devX. Would suggest moving the comment to its own line to emphasize (then slim down)

src/interfaces/IPolicyRegistry.sol
DanielRX: From a CompSci PoV Operands seems okay but low level
From Maths - components / operands 
From Layperson - "Part/Pieces"

"updateCompositeParts/Pieces" feels a lot more approachable if not a bit simple

src/interfaces/IPolicyRegistry.sol
rayyan224: I think we can use policy created. Only thing is I would assume a user want's an event that lists the new childPolicyIds for the composite, when a policy is created / updated.

src/interfaces/IPolicyRegistry.sol
rayyan224: Agreed, I think childPolicyId's give more the intent

src/interfaces/IPolicyRegistry.sol
ilikesymmetry: The existing `PolicyCreated` event does declare the policy type so I think your goal is met as-is

src/interfaces/IPolicyRegistry.sol
```diff
@@ -68,13 +90,26 @@ interface IPolicyRegistry {
     /// @notice One or more accounts had their BLOCKLIST membership set to `blocked` in a single batch.
     event BlocklistUpdated(uint64 indexed policyId, address indexed updater, bool blocked, address[] accounts);
 
+    /// @notice A composite policy (UNION or INTERSECT) was created over `childPolicyIds`.
+    event CompositePolicyCreated(
+        uint64 indexed policyId, address indexed creator, PolicyType policyType, uint64[] childPolicyIds
+    );
```
ilikesymmetry: Seems to ignore discussion here: https://github.com/base/base-std/pull/174#discussion_r3634305482.

As we keep adding policy types with more nuance, we'll continually run up against the desire to have custom created events that have all of the content of `PolicyCreated`, but with a new field(s). I think we should take the opinionated stance that we have one canonical creation event (`PolicyCreated`) and we use `*Updated` events for the type-specific field(s). This gives indexers the cleanest experience where this is only one path for creation to look at ever and only one path for field updates to look at (regardless of creation/update).

```suggestion
```

src/interfaces/IPolicyRegistry.sol
```diff
@@ -154,6 +215,24 @@ interface IPolicyRegistry {
...
+    /// @notice Replaces a composite policy's child-policy set in full with `childPolicyIds`.
+    /// @dev Reverts with `PolicyNotFound` when `policyId` does not exist.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyId` is not a composite (UNION or INTERSECT).
+    /// @dev Reverts with `Unauthorized` when the caller is not the current admin. A renounced composite
+    ///      (admin `address(0)`) can never be updated.
+    /// @dev Reverts with `EmptyChildPolicySet` when `childPolicyIds` is empty; there is no clear-the-list
+    ///      path.
+    /// @dev Reverts with `TooFewChildPolicies` when `childPolicyIds.length == 1`.
+    /// @dev Reverts with `BatchSizeTooLarge` when `childPolicyIds.length` exceeds the composite
+    ///      child-policy limit.
```
ilikesymmetry: How will integrators know what the limit is?

src/interfaces/IPolicyRegistry.sol
rayyan224: But it wouldn't solve the event that the user can see the new childPoloicyId's

src/interfaces/IPolicyRegistry.sol
```diff
@@ -154,6 +215,24 @@ interface IPolicyRegistry {
...
+    /// @dev Reverts with `PolicyNotFound` when any child policy does not exist.
+    /// @dev Reverts with `InvalidChildPolicy` when any child policy is itself a composite
+    ///      (not a simple policy).
```
ilikesymmetry, line 213: EDIT: I agree with the current implementation actually. The nuance of invalid type from invalid type _for a child_ is worth having and the not found error is good to share. Self-resolving

Seeing some inconsistency in design here:
1. We reuse `PolicyNotFound` between children and composite policies
2. We use `IncompatiblePolicyType` for composites and `InvalidChildPolicy` for children

Generally, I would opt for separate errors for both for clarity. That said, I sense that over time we're going to have a massive ABI that's hard to keep track of if we don't reuse where it's possible. Coin toss for me right now.

src/interfaces/IPolicyRegistry.sol
```diff
@@ -154,6 +215,24 @@ interface IPolicyRegistry {
...
+    ///
+    /// @param policyId       Composite policy to update.
+    /// @param childPolicyIds Complete new set of existing simple policy IDs.
+    function updateCompositeChildPolicies(uint64 policyId, uint64[] calldata childPolicyIds) external;
```
ilikesymmetry: Something about this name doesn't pass gut check. If we were consistent with `updateAllowlist`/`updateBlocklist`, we would say `updateComposite`. If we were focused just on the field itself, we'd say `updateChildPolicies` and it's implicit that only composite policies have children policies (for now, as maybe that would change and we'd have some other thing not labeled as "composite" that we think should have children too and would like to reuse the function). Main benefit of this verbose name is that is has consistency with our also-verbose event name.

How do you compare these options?
1. `updateCompositeChildPolicies`
2. `updateComposite`
3. `updateChildPolicies`

ilikesymmetry (changes requested):

src/interfaces/IPolicyRegistry.sol
```diff
@@ -48,6 +56,20 @@ interface IPolicyRegistry {
...
     error NoPendingAdmin();
 
+    /// @notice A composite policy was created or updated with an empty child-policy set.
+    error EmptyChildPolicySet();
+
+    /// @notice A composite policy was created or updated with fewer than the minimum number of
+    ///         child policies. A composite must reference at least two simple policies.
+    /// @param provided Number of child policy IDs supplied.
+    /// @param minimum  Minimum required (2).
+    error TooFewChildPolicies(uint256 provided, uint256 minimum);
```
ilikesymmetry: why do we need empty error if we have the "too few" error?

rayyan224: good point we can combine to use TooFewChildPolicies for case 0

src/interfaces/IPolicyRegistry.sol
rayyan224: For the sake of naming I think updateComposite, the function indicator should indicate that it updates the childPolicies. Will keep it consistent also.

src/interfaces/IPolicyRegistry.sol
rayyan224: So the create function fires two events fires two events, Created and UpdatePolicy ?

src/interfaces/IPolicyRegistry.sol
rayyan224: Re use the `BatchSizeTooLarge( maxBatchSize )` error that we use currently in PolicyRegistry

src/interfaces/IPolicyRegistry.sol
rayyan224: Updated to PolicyCreated, and a UpdateCompositePolicyEvent both being emitted

src/interfaces/IPolicyRegistry.sol
```diff
@@ -48,6 +56,17 @@ interface IPolicyRegistry {
     /// @notice `finalizeUpdateAdmin` was called with no pending admin staged.
     error NoPendingAdmin();
 
+    /// @notice A composite policy was created or updated with fewer than the minimum number of
+    ///         child policies. A composite must reference at least two simple policies.
+    /// @param provided Number of child policy IDs supplied.
+    /// @param minimum  Minimum required (2).
+    error TooFewChildPolicies(uint256 provided, uint256 minimum);
```
stevieraykatz: I don't think this error needs params. The natspec says you need at least two and the user will know how many they provided.

src/interfaces/IPolicyRegistry.sol
```diff
@@ -68,13 +87,19 @@ interface IPolicyRegistry {
...
     /*//////////////////////////////////////////////////////////////
                             POLICY CREATION
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice Creates a new policy with no initial members. Permissionless.
+    /// @notice Creates a new simple policy with no initial members. Permissionless.
     ///
     /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is a composite gate (UNION or
+    ///      INTERSECT); use `createCompositePolicy` for those.
```
stevieraykatz:
```suggestion
    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is a composite gate.
```
I think less is better here

stevieraykatz: Also I think that we use `gate` to refer to composite policyTypes. But only sometimes. I'd probably remove all instances of the word `gate` and instead refer explicitly to `composite policyType`s.

src/interfaces/IPolicyRegistry.sol
```diff
@@ -83,9 +108,11 @@ interface IPolicyRegistry {
     /// @return newPolicyId The newly assigned policy ID.
     function createPolicy(address admin, PolicyType policyType) external returns (uint64 newPolicyId);
 
-    /// @notice Creates a new policy seeded with `accounts` as initial members. Permissionless.
+    /// @notice Creates a new simple policy seeded with `accounts` as initial members. Permissionless.
     ///
     /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`. Takes precedence over `BatchSizeTooLarge`.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is a composite gate (UNION or
+    ///      INTERSECT); use `createCompositePolicy` for those. Takes precedence over `BatchSizeTooLarge`.
```
stevieraykatz:
```suggestion
    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is a composite policyType.
```

cleanup

stevieraykatz: The order of these dev comments should match the revert ordering implicitly and then you don't need explicit revert precedence verbiage in the natspec.

src/interfaces/IPolicyRegistry.sol
```diff
@@ -98,6 +125,29 @@ interface IPolicyRegistry {
         external
         returns (uint64 newPolicyId);
 
+    /// @notice Creates a new composite policy that combines existing simple policies under a logic
+    ///         gate.
+    /// @dev Child policies must be simple policies (ALLOWLIST or BLOCKLIST), never another composite.
```
stevieraykatz, line 129:
```suggestion
    /// @notice Creates a new composite policy that combines existing simple policies under a logic
    ///         gate.
    ///
    /// @dev Child policies must be simple policies (ALLOWLIST or BLOCKLIST), never another composite.
```

src/interfaces/IPolicyRegistry.sol
```diff
@@ -98,6 +125,29 @@ interface IPolicyRegistry {
...
+    ///      The child-policy set is capped at 4.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is not UNION or INTERSECT.
+    /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`. 
+    /// @dev Reverts with `TooFewChildPolicies` when `childPolicyIds.length < 2` (including empty).
+    /// @dev Reverts with `BatchSizeTooLarge(4)` when `childPolicyIds.length > 4`
+    /// @dev Reverts with `PolicyNotFound` when any child policy does not exist.
+    /// @dev Reverts with `InvalidChildPolicy` when any child policy is itself a composite
+    ///      (not a simple policy).
+    /// @dev Panics with arithmetic overflow (Panic 0x11) when the policy counter has reached its maximum value.
```
stevieraykatz, line 136: Aren't we going to explicitly check that `2 < childPolicyIds <= 4`? What risk is there of overflow?

src/interfaces/IPolicyRegistry.sol
```diff
@@ -98,6 +125,29 @@ interface IPolicyRegistry {
...
+    ///         gate.
+    /// @dev Child policies must be simple policies (ALLOWLIST or BLOCKLIST), never another composite.
+    ///      The child-policy set is capped at 4.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is not UNION or INTERSECT.
+    /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`. 
+    /// @dev Reverts with `TooFewChildPolicies` when `childPolicyIds.length < 2` (including empty).
+    /// @dev Reverts with `BatchSizeTooLarge(4)` when `childPolicyIds.length > 4`
+    /// @dev Reverts with `PolicyNotFound` when any child policy does not exist.
+    /// @dev Reverts with `InvalidChildPolicy` when any child policy is itself a composite
+    ///      (not a simple policy).
```
stevieraykatz:
```suggestion
    /// @dev Reverts with `InvalidChildPolicy` when any child policy is not a simple policy.
```

src/interfaces/IPolicyRegistry.sol
```diff
@@ -98,6 +125,29 @@ interface IPolicyRegistry {
         external
         returns (uint64 newPolicyId);
 
+    /// @notice Creates a new composite policy that combines existing simple policies under a logic
+    ///         gate.
+    /// @dev Child policies must be simple policies (ALLOWLIST or BLOCKLIST), never another composite.
+    ///      The child-policy set is capped at 4.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is not UNION or INTERSECT.
+    /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`. 
+    /// @dev Reverts with `TooFewChildPolicies` when `childPolicyIds.length < 2` (including empty).
```
stevieraykatz: empty means 0 length. i think you can remove the parenthetical

src/interfaces/IPolicyRegistry.sol
rayyan224: They share the create  counter which can overflow

src/interfaces/IPolicyRegistry.sol
rayyan224: Is there any, downside of having it ?

stevieraykatz: gas if a reverting tx ever lands onchain

src/interfaces/IPolicyRegistry.sol
stevieraykatz: oooh for the global policy id. i don't think this needs explicit callout here...

src/interfaces/IPolicyRegistry.sol
ilikesymmetry: Ah didn't click for me before, but makes sense. Feels a bit odd given the error is trying to mean something else, but if I reset my expectations I agree it works

src/interfaces/IPolicyRegistry.sol
```diff
@@ -154,6 +199,22 @@ interface IPolicyRegistry {
     /// @param accounts Accounts to update.
     function updateBlocklist(uint64 policyId, bool blocked, address[] calldata accounts) external;
 
+    /// @notice Replaces a composite policy's child-policy set in full with `childPolicyIds`.
+    /// @dev Reverts with `PolicyNotFound` when `policyId` does not exist.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyId` is not a composite (UNION or INTERSECT).
+    /// @dev Reverts with `Unauthorized` when the caller is not the current admin. A renounced composite
+    ///      (admin `address(0)`) can never be updated.
+    /// @dev Reverts with `TooFewChildPolicies` when `childPolicyIds.length < 2`; there is no clear-the-list path.
+    /// @dev Reverts with `BatchSizeTooLarge(4)` when `childPolicyIds.length > 4`
```
ilikesymmetry: asymmetry of these events feels weird, feels like we should do one `OutsideAllowedChildPoliciesRange(received, min, max)` or similar

stephancill (approved):

src/interfaces/IPolicyRegistry.sol
```diff
@@ -68,24 +88,29 @@ interface IPolicyRegistry {
...
     /*//////////////////////////////////////////////////////////////
                             POLICY CREATION
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice Creates a new policy with no initial members. Permissionless.
+    /// @notice Creates a new simple policy with no initial members. Permissionless.
     ///
     /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`.
-    /// @dev Panics with arithmetic overflow (Panic 0x11) when the policy counter has reached its maximum value.
+    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is a composite gate.
```
robriks, line 102: `createPolicy` and `createPolicyWithAccounts` have natspec mentioning that they revert for composite policy types, but it looks like

```
policyRegistry.createPolicy(admin, PolicyType.UNION);      // does not revert, creating a UNION with 0 children
policyRegistry.createPolicy(admin, PolicyType.INTERSECT);  // isAuthorized(id, anyAddr) == true 
policyRegistry.createPolicyWithAccounts(admin, PolicyType.UNION, accts); // does not revert
```

don't revert as documented

rayyan224: Oh yep nice catch

PR #176, opened 2026-07-24
fix: reject composite policyType in createPolicy / createPolicyWithAccounts

rayyan224:
## Summary

Addresses @robriks' review comment on #174 (https://github.com/base/base-std/pull/174#discussion_r3648107360).

The `IPolicyRegistry` natspec states that `createPolicy` and `createPolicyWithAccounts` revert `IncompatiblePolicyType` when `policyType` is a composite gate, but `MockPolicyRegistry` never enforced it:

```solidity
policyRegistry.createPolicy(admin, PolicyType.UNION);      // created a 0-child UNION instead of reverting
policyRegistry.createPolicy(admin, PolicyType.INTERSECT);  // isAuthorized(id, anyAddr) == true (vacuous AND)
policyRegistry.createPolicyWithAccounts(admin, PolicyType.UNION, accts); // did not revert
```

A 0-child composite is degenerate: `isAuthorized` folds an empty child set (UNION → false, INTERSECT → true), silently minting an always-deny / always-authorize policy.

## Changes

- **`MockPolicyRegistry`**: `createPolicy` and `createPolicyWithAccounts` now revert `IncompatiblePolicyType` for UNION/INTERSECT, matching the documented precedence — after the zero-admin check, and (for `createPolicyWithAccounts`) before the batch-size check. Factored the type test into a shared `_isCompositeType(PolicyType)` helper (reused by `createCompositePolicy` and `_isComposite`).
- **Tests**: added `test_createPolicy_revert_incompatiblePolicyType` and `test_createPolicyWithAccounts_revert_incompatiblePolicyType`, and extended both `*_revertOrder` walks with the new step.

Composites remain creatable only through `createCompositePolicy`, which requires a UNION/INTERSECT gate and a `[2, 4]` child set.

## Testing

`forge test` → 661 passed, 0 failed, 4 skipped. `python3 script/check-coverage.py` → all covered (exit 0). `forge fmt --check` clean.

### DIFF
--- a/test/lib/mocks/MockPolicyRegistry.sol
+++ b/test/lib/mocks/MockPolicyRegistry.sol
@@ -97,6 +97,10 @@ contract MockPolicyRegistry is IPolicyRegistry {
 
     /// @inheritdoc IPolicyRegistry
     function createPolicy(address admin, PolicyType policyType) external returns (uint64 newPolicyId) {
+        // ZeroAddress precedes IncompatiblePolicyType (see interface natspec). `_create`
+        // re-checks zero-admin, but the hoisted copy pins the precedence.
+        if (admin == address(0)) revert ZeroAddress();
+        if (_isCompositeType(policyType)) revert IncompatiblePolicyType();
         newPolicyId = _create(admin, policyType);
     }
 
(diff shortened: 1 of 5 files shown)


robriks (approved):
thanks for getting this in!

robriks (approved):

PR #177, opened 2026-07-27
feat(BOP-440): add transferFromSeizableWithMemo and burnBlockedWithMemo seize surface

stephancill:
## Summary

Adds the **seize** operation class to the base-std reference + ABI (the source of truth the fork harness cross-validates against base/base), per the [BOP-439](https://linear.app/coinbase/issue/BOP-439) PPS. Implements [BOP-440](https://linear.app/coinbase/issue/BOP-440) and, per review, also folds in the [BOP-471](https://linear.app/coinbase/issue/BOP-471) `burnBlocked` repoint (base-std portion).

New surface:
- `transferFromSeizableWithMemo(address from, address to, uint256 amount, bytes32 memo)` — admin seize that reassigns a blocked account's balance in one call. Skips allowance and the transfer policies; the only membership check is that `from` is blocked under the new `SEIZABLE_ACCOUNT_POLICY`. `to` is not policy-checked (admin op). Emits `Transfer` + `Memo` + `TransferredFromSeizable`.
- `burnBlockedWithMemo(address from, uint256 amount, bytes32 memo)` — memo variant of burn-based seizure.
- New `SEIZABLE_ACCOUNT_POLICY` scope, `TRANSFER_FROM_SEIZABLE_ROLE`, and `SEIZE` `PausableFeature`.

Seize operation class (all three gate uniformly on `SEIZABLE_ACCOUNT_POLICY` + the `SEIZE` pause vector):
- **`burnBlocked` repointed** from `TRANSFER_SENDER_POLICY`/`BURN` pause to `SEIZABLE_ACCOUNT_POLICY`/`SEIZE` pause. This is a **breaking change**, acceptable because `burnBlocked` is not yet in use; intent is a single uniform seize gate rather than a transient split.

## Design notes

- **Storage: dedicated slot at offset 14.** The seizable id lives in its own `SeizePolicyIds` slot placed **before** the mock-only `initialized` flag (which stays trailing at offset 15). This matters for the Rust mirror: the native precompile checks factory-init via deployed marker bytecode and stores no `initialized` field, so it omits that trailing mock-only slot and lands `seizable_policy_id` at offset 14 — with no filler on either side. Offsets 0–13 are untouched, so already-deployed tokens are undisturbed. Seize is a distinct per-operation group (cold-path, spanning transfer + burn), mirroring how mint policies are grouped separately from transfer policies.
- **Explicit guards, no privileged reuse.** Seize applies its own guard sequence (`SEIZE` pause → role → `SEIZABLE_ACCOUNT_POLICY` block check) and moves balance via a shared `_moveBalance` primitive extracted from `transfer_inner`; it does not reuse the factory-bootstrap `privileged` path.
- **Memo ordering.** `Memo` is emitted immediately after `Transfer` (before `TransferredFromSeizable` / `BurnedBlocked`), per the IB20 `Memo` invariant.

## Testing

- `forge test` (non-fork): all pass (715+), `forge fmt --check` clean.
- Existing `burnBlocked` tests updated to the seize gating; new happy-path + revert-order suites for both new functions.
- Storage layout pinned via `B20FullLayout` (seize at slot 14, `initialized` at slot 15) and `MockB20SlotHelpers` round-trip tests; `B20Removals` regression updated for the new `SEIZE` feature (`ALL_FEATURES_PAUSED` 7 → 15, enum boundary index 4).

### DIFF
--- a/src/interfaces/IB20.sol
+++ b/src/interfaces/IB20.sol
@@ -16,7 +16,7 @@ interface IB20 {
     /// @param TRANSFER `transfer`, `transferFrom`, and memo'd variants.
     /// @param MINT     `mint` and `mintWithMemo`.
     /// @param BURN     `burn`, `burnWithMemo`, and `burnBlocked`.
-    /// @param SEIZE    `transferFromBlockedWithMemo` and `burnBlockedWithMemo` (the seize operation class).
+    /// @param SEIZE    `transferFromBlockedWithMemo` and `burnBlockedWithMemo`
     enum PausableFeature {
         TRANSFER,
         MINT,
@@ -450,7 +450,7 @@ interface IB20 {
     ///
     /// @param from   Account whose balance is being seized.
     /// @param amount Amount to burn.
-    /// @param memo   Off-chain memo payload.
+    /// @param memo   Memo payload.
     function burnBlockedWithMemo(address from, uint256 amount, bytes32 memo) external;
 
     /// @notice Seizes `amount` of `from`'s balance and reassigns it to `to` in a single admin operation.
@@ -470,7 +470,7 @@ interface IB20 {
     /// @param from   Account whose balance is being seized.
     /// @param to     Destination address for the seized balance.
     /// @param amount Amount to seize.
-    /// @param memo   Off-chain memo payload.
+    /// @param memo   Memo payload.
     ///
     /// @return Always `true` on success.
     function transferFromBlockedWithMemo(address from, address to, uint256 amount, bytes32 memo) external returns (bool);
--- a/src/interfaces/IB20.sol
+++ b/src/interfaces/IB20.sol
@@ -16,7 +16,7 @@ interface IB20 {
     /// @param TRANSFER `transfer`, `transferFrom`, and memo'd variants.
     /// @param MINT     `mint` and `mintWithMemo`.
     /// @param BURN     `burn` and `burnWithMemo`.
-    /// @param SEIZE    `burnBlocked`, `burnBlockedWithMemo`, and `transferFromBlockedWithMemo`
+    /// @param SEIZE    `burnBlocked`, `burnBlockedWithMemo`, and `transferFromSeizableWithMemo`
     enum PausableFeature {
         TRANSFER,
         MINT,
@@ -105,8 +105,8 @@ interface IB20 {
     /// @notice `policyScope` is not a slot this token (or its variant) supports.
     error UnsupportedPolicyType(bytes32 policyScope);
 
-    /// @notice A seize operation (`burnBlocked`, `burnBlockedWithMemo`, or `transferFromBlockedWithMemo`) was
-    ///         called against a `from` that is currently authorized under `SEIZABLE_POLICY` (i.e. not blocked).
+    /// @notice A seize operation (`burnBlocked`, `burnBlockedWithMemo`, or `transferFromSeizableWithMemo`) was
+    ///         called against a `from` that is currently authorized under `SEIZABLE_ACCOUNT_POLICY` (i.e. not blocked).
     error AccountNotBlocked(address account);
 
     /// @notice An EIP-2612 `permit` was submitted with a `deadline` strictly less than `block.timestamp`.
@@ -145,10 +145,10 @@ interface IB20 {
     /// @notice Emitted by `burnBlocked` and `burnBlockedWithMemo` in addition to `Transfer(from, address(0), amount)`.
     event BurnedBlocked(address indexed caller, address indexed from, uint256 amount);
 
-    /// @notice Emitted by `transferFromBlockedWithMemo` in addition to `Transfer(from, to, amount)` (and the
+    /// @notice Emitted by `transferFromSeizableWithMemo` in addition to `Transfer(from, to, amount)` (and the
     ///         standard `Memo(caller, memo)`). Records a transfer-based seizure: `caller` is the `msg.sender`
-    ///         (holder of `TRANSFER_FROM_BLOCKED_ROLE`), `from` the seized account, `to` the destination.
-    event TransferredFromBlocked(address indexed caller, address indexed from, address indexed to, uint256 amount);
+    ///         (holder of `TRANSFER_FROM_SEIZABLE_ROLE`), `from` the seized account, `to` the destination.
+    event TransferredFromSeizable(address indexed caller, address indexed from, address indexed to, uint256 amount);
 
     /// @notice Emitted when `account` is granted `role`. `sender` is the originating caller.
     event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender);
@@ -211,9 +211,9 @@ interface IB20 {
     /// @return Role constant.
     function BURN_BLOCKED_ROLE() external view returns (bytes32);
 
-    /// @notice Required to call `transferFromBlockedWithMemo`.
+    /// @notice Required to call `transferFromSeizableWithMemo`.
     /// @return Role constant.
-    function TRANSFER_FROM_BLOCKED_ROLE() external view returns (bytes32);
+    function TRANSFER_FROM_SEIZABLE_ROLE() external view returns (bytes32);
 
     /// @notice Required to call `pause`.
     /// @return Role constant.
@@ -257,12 +257,12 @@ interface IB20 {
     /// @return Policy scope constant.
     function MINT_RECEIVER_POLICY() external view returns (bytes32);
 
-    /// @notice Policy slot consulted against `from` by the seize operations (`transferFromBlockedWithMemo`,
+    /// @notice Policy slot consulted against `from` by the seize operations (`transferFromSeizableWithMemo`,
     ///         `burnBlockedWithMemo`). A `from` is seizable only when it is NOT authorized by this policy
     ///         (mirroring the `burnBlocked` "blocked" semantics). An unset slot reads as `0` (always-allow),
     ///         which makes no account seizable until an issuer configures the slot.
     /// @return Policy scope constant.
-    function SEIZABLE_POLICY() external view returns (bytes32);
+    function SEIZABLE_ACCOUNT_POLICY() external view returns (bytes32);
 
     /*//////////////////////////////////////////////////////////////
                                   ERC-20
@@ -432,7 +432,7 @@ interface IB20 {
     ///
     /// @dev Reverts with `ContractPaused(SEIZE)` when `SEIZE` is paused.
     /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `BURN_BLOCKED_ROLE`.
-    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `SEIZABLE_POLICY`.
+    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `SEIZABLE_ACCOUNT_POLICY`.
     /// @dev Reverts with `InsufficientBalance` when `from`'s balance is below `amount`.
     ///
     /// @param from   Account whose balance is being seized.
@@ -444,7 +444,7 @@ interface IB20 {
     ///
     /// @dev Reverts with `ContractPaused(SEIZE)` when `SEIZE` is paused.
     /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `BURN_BLOCKED_ROLE`.
-    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `SEIZABLE_POLICY`.
+    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `SEIZABLE_ACCOUNT_POLICY`.
     /// @dev Reverts with `InsufficientBalance` when `from`'s balance is below `amount`.
     ///
     /// @param from   Account whose balance is being seized.
@@ -454,15 +454,15 @@ interface IB20 {
 
     /// @notice Seizes `amount` of `from`'s balance and reassigns it to `to` in a single admin operation.
     ///         Emits, in order, `Transfer(from, to, amount)`, `Memo(caller, memo)`, and
-    ///         `TransferredFromBlocked(caller, from, to, amount)`. A memo of `bytes32(0)` is permitted.
+    ///         `TransferredFromSeizable(caller, from, to, amount)`. A memo of `bytes32(0)` is permitted.
     ///
     /// @dev Admin operation: skips allowance and the transfer policies. The only membership check is that
-    ///      `from` is blocked under `SEIZABLE_POLICY`.
+    ///      `from` is blocked under `SEIZABLE_ACCOUNT_POLICY`.
     /// @dev `to` is not policy-checked; the destination need not be allowlisted.
     /// @dev Reverts with `ContractPaused(SEIZE)` when `SEIZE` is paused.
-    /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `TRANSFER_FROM_BLOCKED_ROLE`.
+    /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `TRANSFER_FROM_SEIZABLE_ROLE`.
     /// @dev Reverts with `InvalidReceiver` when `to == address(0)`.
-    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `SEIZABLE_POLICY`.
+    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `SEIZABLE_ACCOUNT_POLICY`.
     /// @dev Reverts with `InsufficientBalance` when `from`'s balance is below `amount`.
     ///
     /// @param from   Account whose balance is being seized.
@@ -471,7 +471,9 @@ interface IB20 {
     /// @param memo   Memo payload.
     ///
     /// @return Always `true` on success.
-    function transferFromBlockedWithMemo(address from, address to, uint256 amount, bytes32 memo) external returns (bool);
+    function transferFromSeizableWithMemo(address from, address to, uint256 amount, bytes32 memo)
+        external
+        returns (bool);
 
     /*//////////////////////////////////////////////////////////////
                                   ROLES
--- a/src/interfaces/IB20.sol
+++ b/src/interfaces/IB20.sol
@@ -441,7 +441,7 @@ interface IB20 {
     function burnBlocked(address from, uint256 amount) external;
 
     /// @notice Same as `burnBlocked`, plus emits `Memo` after the standard `Transfer` and `BurnedBlocked` events.
-    ///         A memo of `bytes32(0)` is permitted. Part of the seize operation class.
+    ///         A memo of `bytes32(0)` is permitted.
     ///
     /// @dev Reverts with `ContractPaused(SEIZE)` when `SEIZE` is paused.
     /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `BURN_BLOCKED_ROLE`.
--- a/src/interfaces/IB20.sol
+++ b/src/interfaces/IB20.sol
@@ -15,8 +15,8 @@ interface IB20 {
     ///
     /// @param TRANSFER `transfer`, `transferFrom`, and memo'd variants.
     /// @param MINT     `mint` and `mintWithMemo`.
-    /// @param BURN     `burn`, `burnWithMemo`, and `burnBlocked`.
-    /// @param SEIZE    `transferFromBlockedWithMemo` and `burnBlockedWithMemo`
+    /// @param BURN     `burn` and `burnWithMemo`.
+    /// @param SEIZE    `burnBlocked`, `burnBlockedWithMemo`, and `transferFromBlockedWithMemo`
     enum PausableFeature {
         TRANSFER,
         MINT,
@@ -105,9 +105,8 @@ interface IB20 {
     /// @notice `policyScope` is not a slot this token (or its variant) supports.
     error UnsupportedPolicyType(bytes32 policyScope);
 
-    /// @notice A seize operation was called against a `from` that is not currently blocked by the gating policy:
-    ///         `burnBlocked` (authorized under `TRANSFER_SENDER_POLICY`), or `transferFromBlockedWithMemo` /
-    ///         `burnBlockedWithMemo` (authorized under `SEIZABLE_POLICY`).
+    /// @notice A seize operation (`burnBlocked`, `burnBlockedWithMemo`, or `transferFromBlockedWithMemo`) was
+    ///         called against a `from` that is currently authorized under `SEIZABLE_POLICY` (i.e. not blocked).
     error AccountNotBlocked(address account);
 
     /// @notice An EIP-2612 `permit` was submitted with a `deadline` strictly less than `block.timestamp`.
@@ -429,11 +428,11 @@ interface IB20 {
     function burnWithMemo(uint256 amount, bytes32 memo) external;
 
     /// @notice Destroys `amount` of `from`'s balance. Emits `Transfer(from, address(0), amount)` and
-    ///         `BurnedBlocked(caller, from, amount)`.
+    ///         `BurnedBlocked(caller, from, amount)`. Part of the seize operation class.
     ///
-    /// @dev Reverts with `ContractPaused(BURN)` when `BURN` is paused.
+    /// @dev Reverts with `ContractPaused(SEIZE)` when `SEIZE` is paused.
     /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `BURN_BLOCKED_ROLE`.
-    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `TRANSFER_SENDER_POLICY`.
+    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `SEIZABLE_POLICY`.
     /// @dev Reverts with `InsufficientBalance` when `from`'s balance is below `amount`.
     ///
     /// @param from   Account whose balance is being seized.
--- a/src/interfaces/IB20.sol
+++ b/src/interfaces/IB20.sol
@@ -16,10 +16,12 @@ interface IB20 {
     /// @param TRANSFER `transfer`, `transferFrom`, and memo'd variants.
     /// @param MINT     `mint` and `mintWithMemo`.
     /// @param BURN     `burn`, `burnWithMemo`, and `burnBlocked`.
+    /// @param SEIZE    `transferFromBlockedWithMemo` and `burnBlockedWithMemo` (the seize operation class).
     enum PausableFeature {
         TRANSFER,
         MINT,
-        BURN
+        BURN,
+        SEIZE
     }
 
     /*//////////////////////////////////////////////////////////////
@@ -103,7 +105,9 @@ interface IB20 {
     /// @notice `policyScope` is not a slot this token (or its variant) supports.
     error UnsupportedPolicyType(bytes32 policyScope);
 
-    /// @notice `burnBlocked` was called against a `from` that is currently authorized under `TRANSFER_SENDER_POLICY`.
+    /// @notice A seize operation was called against a `from` that is not currently blocked by the gating policy:
+    ///         `burnBlocked` (authorized under `TRANSFER_SENDER_POLICY`), or `transferFromBlockedWithMemo` /
+    ///         `burnBlockedWithMemo` (authorized under `SEIZABLE_POLICY`).
     error AccountNotBlocked(address account);
 
     /// @notice An EIP-2612 `permit` was submitted with a `deadline` strictly less than `block.timestamp`.
@@ -139,8 +143,13 @@ interface IB20 {
     ///         immediately after the underlying `Transfer` event. `caller` is the `msg.sender` of the memo'd call.
     event Memo(address indexed caller, bytes32 indexed memo);
 
-    /// @notice Emitted by `burnBlocked` in addition to `Transfer(from, address(0), amount)`.
+    /// @notice Emitted by `burnBlocked` and `burnBlockedWithMemo` in addition to `Transfer(from, address(0), amount)`.
     event BurnedBlocked(address indexed caller, address indexed from, uint256 amount);
+
+    /// @notice Emitted by `transferFromBlockedWithMemo` in addition to `Transfer(from, to, amount)` (and the
+    ///         standard `Memo(caller, memo)`). Records a transfer-based seizure: `caller` is the `msg.sender`
+    ///         (holder of `TRANSFER_FROM_BLOCKED_ROLE`), `from` the seized account, `to` the destination.
+    event TransferredFromBlocked(address indexed caller, address indexed from, address indexed to, uint256 amount);
 
     /// @notice Emitted when `account` is granted `role`. `sender` is the originating caller.
     event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender);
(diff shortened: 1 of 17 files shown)


src/interfaces/IB20.sol
```diff
@@ -16,10 +16,12 @@ interface IB20 {
     /// @param TRANSFER `transfer`, `transferFrom`, and memo'd variants.
     /// @param MINT     `mint` and `mintWithMemo`.
     /// @param BURN     `burn`, `burnWithMemo`, and `burnBlocked`.
+    /// @param SEIZE    `transferFromBlockedWithMemo` and `burnBlockedWithMemo` (the seize operation class).
```
stephancill:
```suggestion
    /// @param SEIZE    `transferFromBlockedWithMemo` and `burnBlockedWithMemo`
```

src/interfaces/IB20.sol
```diff
@@ -420,6 +440,41 @@ interface IB20 {
...
+    ///         A memo of `bytes32(0)` is permitted. Part of the seize operation class.
+    ///
+    /// @dev Reverts with `ContractPaused(SEIZE)` when `SEIZE` is paused.
+    /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `BURN_BLOCKED_ROLE`.
+    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `SEIZABLE_POLICY`.
+    /// @dev Reverts with `InsufficientBalance` when `from`'s balance is below `amount`.
+    ///
+    /// @param from   Account whose balance is being seized.
+    /// @param amount Amount to burn.
+    /// @param memo   Off-chain memo payload.
```
stephancill: is it necessary to specify offchain here?

src/interfaces/IB20.sol
```diff
@@ -420,6 +440,41 @@ interface IB20 {
...
+    /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `TRANSFER_FROM_BLOCKED_ROLE`.
+    /// @dev Reverts with `InvalidReceiver` when `to == address(0)`.
+    /// @dev Reverts with `InvalidSender` when `from == address(0)`.
+    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `SEIZABLE_POLICY`.
+    /// @dev Reverts with `InsufficientBalance` when `from`'s balance is below `amount`.
+    ///
+    /// @param from   Account whose balance is being seized.
+    /// @param to     Destination address for the seized balance.
+    /// @param amount Amount to seize.
+    /// @param memo   Off-chain memo payload.
```
stephancill: here too

src/lib/B20Constants.sol
```diff
@@ -17,9 +18,10 @@ library B20Constants {
     bytes32 internal constant TRANSFER_RECEIVER_POLICY = keccak256("TRANSFER_RECEIVER_POLICY");
     bytes32 internal constant TRANSFER_EXECUTOR_POLICY = keccak256("TRANSFER_EXECUTOR_POLICY");
     bytes32 internal constant MINT_RECEIVER_POLICY = keccak256("MINT_RECEIVER_POLICY");
+    bytes32 internal constant SEIZABLE_POLICY = keccak256("SEIZABLE_POLICY");
 
-    /// @notice Bitmask with all `PausableFeature` bits set (TRANSFER | MINT | BURN).
-    uint8 internal constant ALL_FEATURES_PAUSED = 7;
+    /// @notice Bitmask with all `PausableFeature` bits set (TRANSFER | MINT | BURN | SEIZE).
+    uint8 internal constant ALL_FEATURES_PAUSED = 15;
```
stephancill: explain this change

test/lib/mocks/MockB20.sol
```diff
@@ -737,6 +774,16 @@ abstract contract MockB20 is IB20 {
...
+        _moveBalance(from, to, amount);
+    }
+
+    /// @dev Shared balance mover: debit `from`, credit `to`, emit `Transfer`.
+    ///      No policy, allowance, or pause checks — callers (`_transfer`,
+    ///      `transferFromBlockedWithMemo`) apply their own guards first. This
+    ///      is the single balance-move primitive so the seize path does not
+    ///      have to reuse the policy-bearing `_transfer` (nor the factory
+    ///      privileged bypass).
+    function _moveBalance(address from, address to, uint256 amount) internal {
```
stephancill, line 782: unsure if it's necessary to be so verbose here - @stevieraykatz wdyt?

test/lib/mocks/MockB20Storage.sol
```diff
@@ -49,17 +49,20 @@ library MockB20Storage {
...
+    /// @notice Transfer-side + seize policy IDs (read by `_transfer`,
+    ///         `transferFrom*`, the seize check in `burnBlocked`, and the
+    ///         seize operations `transferFromBlockedWithMemo` /
+    ///         `burnBlockedWithMemo`).
     /// @dev    Bit layout (Solidity LSB-first):
     ///           bits   0.. 63 : sender
     ///           bits  64..127 : receiver
     ///           bits 128..191 : executor
-    ///           bits 192..255 : reserved (implicit, no field declared)
+    ///           bits 192..255 : seizable (`SEIZABLE_POLICY`)
```
stephancill: does adding seizable here use up the reserved bits? no more reserved bits after adding it?

stephancill: AI: Yes — this fills the slot. `TransferPolicyIds` is one 256-bit slot = 4 × uint64 lanes; sender/receiver/executor used lanes 0–2, and `seizable` takes lane 3 (bits 192..255), which was the last reserved lane. No reserved bits remain in this slot. A future transfer-side policy would need a new appended slot. I picked this lane specifically because it avoids shifting any existing field offset on the already-shipped layout.

src/lib/B20Constants.sol
stephancill: AI: `ALL_FEATURES_PAUSED` is the bitmask with every `PausableFeature` bit set. Adding `SEIZE` as the 4th feature (enum index 3 → bit 3) moves the all-set mask from `0b111` (7) to `0b1111` (15). I added an inline note about the bump, and updated the `B20Removals` regression test (which intentionally pins this value) to 15, plus its enum-boundary check (index 4 is now the out-of-range one).

src/interfaces/IB20.sol
stephancill: AI: Good call — dropped "Off-chain". Per the design thread the memo can be a transparent on-chain message OR a hash of off-chain data, so "off-chain" was too restrictive. Changed both new functions to just "Memo payload." The existing `*WithMemo` natspec still says "Off-chain memo payload." — happy to align those in a follow-up for consistency if you'd like.

src/interfaces/IB20.sol
stephancill: AI: Same fix here — changed to "Memo payload." (see the reply on `burnBlockedWithMemo`).

test/lib/mocks/MockB20.sol
stephancill: AI: Trimmed to two lines (no guards; callers guard first). @stevieraykatz let me know if you'd prefer it shorter or with more detail.

src/lib/B20Constants.sol
```diff
@@ -20,7 +20,8 @@ library B20Constants {
     bytes32 internal constant MINT_RECEIVER_POLICY = keccak256("MINT_RECEIVER_POLICY");
     bytes32 internal constant SEIZABLE_POLICY = keccak256("SEIZABLE_POLICY");
 
-    /// @notice Bitmask with all `PausableFeature` bits set (TRANSFER | MINT | BURN | SEIZE).
+    /// @notice Bitmask with all `PausableFeature` bits set (TRANSFER | MINT | BURN | SEIZE = 0b1111).
+    ///         Raised from `7` (0b111) when the `SEIZE` feature was added as the fourth pausable class.
```
stephancill: we don't need this explanatory note - can we use 0b111 instead of the decimal number in the actual constant declaration for clarity?

stephancill: AI: Dropped the note. Heads up: Solidity has no binary literals (`0b1111` fails to compile — "Identifier-start is not allowed at end of a number"), so I used hex `0x0F` (= 0b1111) as the closest clear bitmask form. Let me know if you'd rather keep plain `15`.

stephancill: keep 15 and add a note about 15 = 0b1111

stephancill: AI: Done — reverted to `15` with an inline `15 = 0b1111` note.

test/lib/mocks/MockB20.sol
stevieraykatz: Method and verbiage seem fine. OZ's ERC20 calls their analog method `_update()`. mIght be worth considering their precedent.

test/lib/mocks/MockB20Storage.sol
stevieraykatz: These bits were reserved for exactly this kind of purpose. Good to flag that we won't have any headroom after this lands, but it's correct to consume them like this.

src/interfaces/IB20.sol
```diff
@@ -420,6 +440,41 @@ interface IB20 {
     /// @param amount Amount to burn.
     function burnBlocked(address from, uint256 amount) external;
 
+    /// @notice Same as `burnBlocked`, plus emits `Memo` after the standard `Transfer` and `BurnedBlocked` events.
+    ///         A memo of `bytes32(0)` is permitted. Part of the seize operation class.
```
```suggestion
    /// @notice Same as `burnBlocked`, plus emits `Memo` after the standard `Transfer` and `BurnedBlocked` events.
    ///         A memo of `bytes32(0)` is permitted.
```

test/lib/mocks/MockB20.sol
stephancill: Got AI to look into it, seems like renaming could be a bit misleading

AI: Good call-out. I opted to keep _moveBalance rather than _update for two reasons:
1. Scope mismatch with OZ's _update. OZ's _update(from, to, value) is the single chokepoint for all balance changes — mint (from == 0), burn (to == 0), and transfer. Here it's transfer-only: mint goes through _mint and burn through _burnRaw, each with their own supply/policy logic. Naming this _update would imply it's the universal hook when it deliberately isn't, which I think would be more confusing than clarifying.
2. Parity with the precompile. This mock mirrors the Rust B-20 precompile, where the same primitive is named move_balance. Keeping _moveBalance preserves the 1:1 base-std ↔ base/base mapping; _update would break that.

src/interfaces/IB20.sol
```diff
@@ -420,6 +440,41 @@ interface IB20 {
...
+    /// @param memo   Memo payload.
+    function burnBlockedWithMemo(address from, uint256 amount, bytes32 memo) external;
+
+    /// @notice Seizes `amount` of `from`'s balance and reassigns it to `to` in a single admin operation.
+    ///         Emits `Transfer(from, to, amount)`, `TransferredFromBlocked(caller, from, to, amount)`, and
+    ///         `Memo(caller, memo)`. A memo of `bytes32(0)` is permitted.
+    ///
+    /// @dev Admin operation: skips allowance and does NOT consult `TRANSFER_SENDER_POLICY`,
+    ///      `TRANSFER_RECEIVER_POLICY`, or `TRANSFER_EXECUTOR_POLICY`. The only membership check is that `from`
+    ///      is blocked under `SEIZABLE_POLICY`; `to` is not policy-checked (the destination need not be allowlisted).
```
stevieraykatz: Tighten this one up a bit. Not sure we need to enumerate the policies we don't consult... and the `to` policy statement might want to be its own note.

src/interfaces/IB20.sol
```diff
@@ -16,10 +16,12 @@ interface IB20 {
     /// @param TRANSFER `transfer`, `transferFrom`, and memo'd variants.
     /// @param MINT     `mint` and `mintWithMemo`.
     /// @param BURN     `burn`, `burnWithMemo`, and `burnBlocked`.
+    /// @param SEIZE    `transferFromBlockedWithMemo` and `burnBlockedWithMemo`
     enum PausableFeature {
         TRANSFER,
         MINT,
-        BURN
+        BURN,
+        SEIZE
```
stevieraykatz, line 24: Why do we need to be able to pause seizing? Seems like the kind of admin action you don't want to stop.

src/interfaces/IB20.sol
```diff
@@ -103,7 +105,9 @@ interface IB20 {
     /// @notice `policyScope` is not a slot this token (or its variant) supports.
     error UnsupportedPolicyType(bytes32 policyScope);
 
-    /// @notice `burnBlocked` was called against a `from` that is currently authorized under `TRANSFER_SENDER_POLICY`.
+    /// @notice A seize operation was called against a `from` that is not currently blocked by the gating policy:
+    ///         `burnBlocked` (authorized under `TRANSFER_SENDER_POLICY`), or `transferFromBlockedWithMemo` /
+    ///         `burnBlockedWithMemo` (authorized under `SEIZABLE_POLICY`).
```
stevieraykatz: It feels wrong to me that these don't share a policy. I know its because we've got BC to worry about, but it still smells. Is this discussed elsewhere?

test/lib/mocks/MockB20.sol
stevieraykatz: precompile parity is compelling.

src/lib/B20Constants.sol
```diff
@@ -17,9 +18,10 @@ library B20Constants {
     bytes32 internal constant TRANSFER_RECEIVER_POLICY = keccak256("TRANSFER_RECEIVER_POLICY");
     bytes32 internal constant TRANSFER_EXECUTOR_POLICY = keccak256("TRANSFER_EXECUTOR_POLICY");
     bytes32 internal constant MINT_RECEIVER_POLICY = keccak256("MINT_RECEIVER_POLICY");
+    bytes32 internal constant SEIZABLE_POLICY = keccak256("SEIZABLE_POLICY");
```
stephancill: @stevieraykatz thoughts on the naming here? an alternative that was floated was `SEIZE_BLOCKED_POLICY`

src/interfaces/IB20.sol
stephancill: the intention here was that you should be able to pause the other functions of the contract without pausing seizing. one option for implementing that is to not make seizing pausable at all, the other one is this one where seizing becomes its own pause vector

src/interfaces/IB20.sol
stephancill: i think this is actually a mistake - we are willing to break BC for burnBlocked under the assumption that it isn't being used yet. intention is to migrate it to the same policy

stephancill: ah i see what happened - that breaking change diff was slated for a separate PR. gonna fold it into this one

src/lib/B20Constants.sol
stevieraykatz: Thanks for drawing my attention. The verbiage is strange but `seizable` is actually accurately following convention. We want to describe what the policy does by describing who the policy applies to. So since we're talking about the holders who can be seized from, `SEIZABLE` is fine. 

I'm wondering if there's a way to incorporate the holder though? `SEIZEABLE_HOLDER_POLICY` doesn't feel right but I also don't like that we don't name the account.

src/interfaces/IB20.sol
stevieraykatz: given `transferFromBlockedWithMemo` ignores our other safeguards i think it could make sense that we would also ignore pause state. 

I would prefer us to know that we need to be able to pause seizing action rather than implement something to avoid its use.

stephancill: I think that makes sense

@ilikesymmetry mind weighing in here?

src/lib/B20Constants.sol
stephancill: `SEIZEABLE_HOLDER_POLICY` could work. Another option could be `SEIZEABLE_ACCOUNT_POLICY`?

stevieraykatz: Are we intending for this policy to be an allowlist or a blocklist? I think that might change how we frame the policy name.

src/interfaces/IB20.sol
ilikesymmetry: @stevieraykatz reference pause options and tradeoffs [here](https://docs.google.com/document/d/1teCPhKWLb6jD-6l5fnPZsWqf0sPL_Gi1uLJETdCojaM/edit?tab=t.0#heading=h.4kbo36r0znf4)

We landed on a new pause option to separate concerns. Including an option to pause seizing operations helps asset issuers reduce risk of a buggy/compromised role-holding account by having a separate system able to killswitch the operation.

src/interfaces/IB20.sol
```diff
@@ -245,6 +257,13 @@ interface IB20 {
     /// @return Policy scope constant.
     function MINT_RECEIVER_POLICY() external view returns (bytes32);
 
+    /// @notice Policy slot consulted against `from` by the seize operations (`transferFromBlockedWithMemo`,
+    ///         `burnBlockedWithMemo`). A `from` is seizable only when it is NOT authorized by this policy
+    ///         (mirroring the `burnBlocked` "blocked" semantics). An unset slot reads as `0` (always-allow),
+    ///         which makes no account seizable until an issuer configures the slot.
+    /// @return Policy scope constant.
+    function SEIZABLE_POLICY() external view returns (bytes32);
```
ilikesymmetry: I don't remember aligning on this as the name and recall we had `SEIZE_BLOCKED_POLICY`. What's the thought process of this new name?

src/interfaces/IB20.sol
```diff
@@ -409,17 +428,52 @@ interface IB20 {
...
+    /// @dev Reverts with `InsufficientBalance` when `from`'s balance is below `amount`.
+    ///
+    /// @param from   Account whose balance is being seized.
+    /// @param amount Amount to burn.
+    /// @param memo   Memo payload.
+    function burnBlockedWithMemo(address from, uint256 amount, bytes32 memo) external;
+
+    /// @notice Seizes `amount` of `from`'s balance and reassigns it to `to` in a single admin operation.
+    ///         Emits `Transfer(from, to, amount)`, `TransferredFromBlocked(caller, from, to, amount)`, and
+    ///         `Memo(caller, memo)`. A memo of `bytes32(0)` is permitted.
```
ilikesymmetry: for clarity, should enumerate events in order of emission, so `Memo` would immediately follow `Transfer`

src/interfaces/IB20.sol
```diff
@@ -409,17 +428,52 @@ interface IB20 {
...
     /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `BURN_BLOCKED_ROLE`.
-    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `TRANSFER_SENDER_POLICY`.
+    /// @dev Reverts with `AccountNotBlocked` when `from` is currently authorized under `SEIZABLE_POLICY`.
     /// @dev Reverts with `InsufficientBalance` when `from`'s balance is below `amount`.
     ///
     /// @param from   Account whose balance is being seized.
     /// @param amount Amount to burn.
     function burnBlocked(address from, uint256 amount) external;
 
+    /// @notice Same as `burnBlocked`, plus emits `Memo` after the standard `Transfer` and `BurnedBlocked` events.
```
ilikesymmetry: for clarity, we emit `Memo` immediately after `Transfer` and before `BurnBlocked`

src/lib/B20Constants.sol
ilikesymmetry: > Are we intending for this policy to be an allowlist or a blocklist? I think that might change how we frame the policy name.

It's most likely going to be a blocklist model, but theoretically we can't guarantee how people will use the scope

ilikesymmetry: > I'm wondering if there's a way to incorporate the holder though? `SEIZEABLE_HOLDER_POLICY` doesn't feel right but I also don't like that we don't name the account.

This was the intent of "SEIZE_BLOCKED" where we describe the entity as being seized from as a "blocked" address. It's a weird word to use as a noun, but it does feel consistent with our naming conventions "transfer from a blocked account" -> "blocked" is the noun

test/lib/mocks/MockB20.sol
```diff
@@ -308,24 +310,49 @@ abstract contract MockB20 is IB20 {
...
 
+    function burnBlockedWithMemo(address from, uint256 amount, bytes32 memo)
+        external
+        whenNotPaused(PausableFeature.SEIZE)
+        onlyRole(BURN_BLOCKED_ROLE)
+    {
+        _requireSeizable(from);
+        _burnRaw(from, amount);
+        emit BurnedBlocked(msg.sender, from, amount);
+        emit Memo(msg.sender, memo);
```
ilikesymmetry: Memo always needs to follow `Transfer` per [spec](https://github.com/base/base-std/blob/main/src/interfaces/IB20.sol#L138-L139). @rayyan224 can we get some automated agent workflow going that consistently looks for accidental divergence from previously defined invariants? This one should be caught before it hits human review

test/lib/mocks/MockB20.sol
```diff
@@ -308,24 +310,49 @@ abstract contract MockB20 is IB20 {
...
+        // executor policy is consulted and no allowance is spent. The only
+        // membership check is that `from` is blocked under SEIZABLE_POLICY.
+        // Deliberately does NOT reuse the factory-bootstrap privileged path
+        // (which would silently skip the receiver policy); every skip here is
+        // explicit.
+        _requireNonZeroActors(from, to);
+        _requireSeizable(from);
+        _moveBalance(from, to, amount);
+        emit TransferredFromBlocked(msg.sender, from, to, amount);
+        emit Memo(msg.sender, memo);
```
ilikesymmetry: same comment on memo ordering as burnBlocked

src/interfaces/IB20.sol
stephancill: discussed some options with @rayyan224 async and felt `SEIZABLE_POLICY` was maybe more descriptive so proceeded with it as a placeholder with the intention of discussing here

test/lib/mocks/MockB20.sol
```diff
@@ -308,24 +310,49 @@ abstract contract MockB20 is IB20 {
...
+        returns (bool)
+    {
+        // Admin seize: reassign a blocked account's balance. `to` is validated
+        // for non-zero, but — unlike a normal transfer — no sender/receiver/
+        // executor policy is consulted and no allowance is spent. The only
+        // membership check is that `from` is blocked under SEIZABLE_POLICY.
+        // Deliberately does NOT reuse the factory-bootstrap privileged path
+        // (which would silently skip the receiver policy); every skip here is
+        // explicit.
+        _requireNonZeroActors(from, to);
```
ilikesymmetry: Why do we care to check `from` is address zero? I agree it should not be, but this then motivates if we need to retroactively add this check to burn-blocked functions as well

test/lib/mocks/MockB20Storage.sol
ilikesymmetry: @rayyan224 another candidate for automated agent fuzzing would be any changes to storage layout (including appending)

ilikesymmetry: I'm a bit concerned to take the last space here. Main thing on my mind is if we have some other idea that definitely needs to be on the hotpath transfer. The seizing operations are cold-path and we won't be using any of the other policies here in its execution, so I'm not seeing as clear of a benefit tbh. I'd opt to make a new slot for seize policy entirely by default.

We also recognize seize as an operation applies to burning too. Placing this policy in the "transfer policies" group is not consistent with our separation of "mint policies" as a distinct group by operation.

test/lib/mocks/MockB20.sol
```diff
@@ -480,6 +507,7 @@ abstract contract MockB20 is IB20 {
         if (policyScope == TRANSFER_SENDER_POLICY) return $.transferPolicyIds.sender;
         if (policyScope == TRANSFER_RECEIVER_POLICY) return $.transferPolicyIds.receiver;
         if (policyScope == TRANSFER_EXECUTOR_POLICY) return $.transferPolicyIds.executor;
+        if (policyScope == SEIZABLE_POLICY) return $.transferPolicyIds.seizable;
         if (policyScope == MINT_RECEIVER_POLICY) return $.mintPolicyIds.receiver;
         revert UnsupportedPolicyType(policyScope);
```
ilikesymmetry: I'm concerned that our `UnsupportedPolicyType` revert is overly restrictive. Thinking through the rollout of these new changes, compliant assets live in production will be temporarily stripped of their ability to block users from holding their assets. To re-enable blocking, they'll need to manually queue up a transaction to set the seize policy to whatever blocklist they were previously managing to keep going. Setting ourselves up for a race condition of sequencing Cobalt activation -> update seize policy -> continue blocking+seizing feels like an unnecessary friction.

**Proposal: Consider removing restriction of defined policy types and treat policies more like roles (open-ended with B20-aware values).** If we had this, we could have enabled teams live in production to queue up their seize policy ahead of the Cobalt upgrade and have zero race conditions or risk of service disturbance. We'll still hit this for Cobalt as we can't change this logic until the upgrade goes live, but I think this sets us up better for changes like this in the future.

ilikesymmetry (changes requested):
Seeking alignment on policy name, storage location, and revisiting behavior of only supporting specific policy scopes.

test/lib/mocks/MockB20.sol
ilikesymmetry: Additionally, we should start to temp check this rollout plan with active customers inside Coinbase on stock and wrapped asset products

src/lib/B20Constants.sol
stephancill: > It's most likely going to be a blocklist model, but theoretically we can't guarantee how people will use the scope

This is my main hesitation against `SEIZE_BLOCKED_POLICY`. Adding an address to the seizing policy doesn't enforce any blocking logic

src/interfaces/IB20.sol
stephancill: AI: Fixed in f0660e2 — natspec now lists emission order `Transfer`, `Memo`, `TransferredFromBlocked`, and the impl emits `Memo` immediately after `Transfer` (before `TransferredFromBlocked`) to satisfy the IB20 `Memo` invariant.

src/interfaces/IB20.sol
stephancill: AI: Fixed in f0660e2 — `burnBlockedWithMemo` now emits `Transfer` -> `Memo` -> `BurnedBlocked`, and the natspec reflects that order.

test/lib/mocks/MockB20.sol
stephancill: AI: Fixed in f0660e2 — reordered to `Transfer` -> `Memo` -> `BurnedBlocked` so `Memo` immediately follows `Transfer` per the spec invariant.

test/lib/mocks/MockB20.sol
stephancill: AI: Fixed in f0660e2 — `transferFromBlockedWithMemo` now emits `Transfer` -> `Memo` -> `TransferredFromBlocked`.

test/lib/mocks/MockB20.sol
stephancill: AI: Dropped the `from == address(0)` check in f0660e2 (kept the `to != 0` check, which stops a seize silently becoming a burn). This matches the burn-blocked family, which doesn't zero-check `from` — a zero/empty `from` fails the seizable or balance check anyway. So no need to retroactively add it to the burn-blocked functions.

src/lib/B20Constants.sol
stephancill: This is also making me second guess the `transferFromBlocked` function naming. `seize` or `transferFromSeizable` is perhaps more accurate

src/interfaces/IB20.sol
rayyan224: I think we can do as discussed in the P/PS ? Seems' like it's related to this duscusion here 
https://github.com/base/base-std/pull/177#discussion_r3659884392

src/lib/B20Constants.sol
stevieraykatz: +1 to `transferFromSeizable` and then naming the policy according to whether the account can be seized from, i.e. `SEIZABLE_ACCOUNT_POLICY`

test/lib/mocks/MockB20.sol
rayyan224: Yes, agreed @stevieraykatz we had some encoded smart contract skills, that you were working on right ?

stevieraykatz: Our solidity harness is the shared repo for agent rules. If this skill/rule isn't defined there, it should be added!

https://coinbase.ghe.com/smart-contracts/harness

test/lib/mocks/MockB20Storage.sol
stephancill: aligned with moving to a dedicated slot

test/lib/mocks/MockB20Storage.sol
```diff
@@ -49,19 +49,25 @@ library MockB20Storage {
...
+    ///           bits 192..255 : reserved (implicit, no field declared)
     struct TransferPolicyIds {
         uint64 sender;
         uint64 receiver;
         uint64 executor;
+    }
+
+    /// @notice Seize policy IDs (read by the seize operations `transferFromSeizableWithMemo`,
+    ///         `burnBlocked`, and `burnBlockedWithMemo`). A distinct per-operation group — seize is
+    ///         cold-path and applies to burns too, so it does not belong in the transfer slot.
```
stephancill: remove the "A distinct per-operation group..." note

test/lib/mocks/MockB20.sol
stephancill: @ilikesymmetry wdyt of adding an unsafe variant of `updatePolicy` which skips this check as opposed to removing the error entirely?

stephancill: improving migration devx is arguably a separate concern from this PR

stephancill:
@ilikesymmetry implemented the following

### Naming
policy: `SEIZABLE_ACCOUNT_POLICY`
fn: `transferFromSeizableWithMemo(from, to, amount, memo)`
role: `TRANSFER_FROM_SEIZABLE_ROLE`
event: `TransferredFromSeizable`

### Policy slot
New slot

### Policy scope support behavior
Concern to be investigated separately

robriks (approved):

src/interfaces/IB20.sol
```diff
@@ -245,6 +257,13 @@ interface IB20 {
     /// @return Policy scope constant.
     function MINT_RECEIVER_POLICY() external view returns (bytes32);
 
+    /// @notice Policy slot consulted against `from` by the seize operations (`transferFromSeizableWithMemo`,
+    ///         `burnBlockedWithMemo`). A `from` is seizable only when it is NOT authorized by this policy
+    ///         (mirroring the `burnBlocked` "blocked" semantics). An unset slot reads as `0` (always-allow),
+    ///         which makes no account seizable until an issuer configures the slot.
+    /// @return Policy scope constant.
```
stevieraykatz: Should probably split into a notice and dev block to move the verbosity into the appropriate field.

stevieraykatz: Also missing `burnBlocked` in `@noice` enumeration

test/lib/mocks/MockB20Storage.sol
```diff
@@ -136,14 +144,20 @@ library MockB20Storage {
...
+        // original offset when `seizePolicyIds` was appended after it, so the
+        // Rust layout omits `initialized` by leaving its slot as an unused hole
+        // and placing `seizePolicyIds` at the same offset as here.
         bool initialized;
+        // ---------- Seize policies (PACKED, per-operation) ----------
+        // Appended after `initialized` so no pre-existing field offset shifts
+        // on already-deployed tokens. Distinct per-operation group (seize is
+        // cold-path and spans transfer + burn), mirroring how mint policies are
+        // grouped separately from transfer policies.
+        SeizePolicyIds seizePolicyIds;
```
stevieraykatz, line 151:
```suggestion
        // ---------- Seize policies (PACKED, per-operation) ----------
        // Appended after `initialized` so no pre-existing field offset shifts
        // on already-deployed tokens.
        SeizePolicyIds seizePolicyIds;
```

stevieraykatz (approved):

stephancill:
@ilikesymmetry going to merge - happy to follow up with changes

PR #178, opened 2026-07-28
test(BOP-472): composite policy smoke coverage

rayyan224:
Extends the policy-registry smoke journey from 16 to **35 steps** for composite (UNION/INTERSECT) policies, plus `NonPayable` coverage for the two new selectors in `precompile_invariants.py`.

**Smoke lane only.** The forge unit-test work (33 tests, runtime V2 gating, V1-negative suite) is parked on `rayyanalam/bop-472-union-policy-forge-tests` and will follow as its own PR. Nothing under `test/` changes here.

## New steps

| Steps | Coverage |
|---|---|
| 17–21 | UNION/INTERSECT creation; OR/AND evaluation; live re-evaluation on child membership change; `updateComposite` full-replace + exact event payload |
| 22–25 | child-count range, built-in child, composite child, unauthorized update |
| 26 | composite creator input guards — `ZeroAddress`, `IncompatiblePolicyType`; plus the mirror guards (`updateComposite` on a simple id, membership mutators on a composite id) |
| 27 | non-existent child → `PolicyNotFound`, in both first and last array position, and its precedence over `InvalidChildPolicy` |
| 28–31 | token enforcement through a composite — mint/transfer allowed, `PolicyForbids` denied |
| 32 | flow-level events incl. `CompositePolicyUpdated` |
| 33–35 | base/base conformance checks — **currently RED**, see below |

Steps 26–27 close the last gaps in the original `smokeTests.md` item-4 revert list. `Panic 0x11` (counter at max) is the one item deliberately uncovered — not reachable on a live chain.

## Steps 33–35: base/base does not match the interface

base-std's `IPolicyRegistry` is the contract; the Rust precompile is the implementation of it. **There are no accepted divergences** — every check here is a base/base bug to fix, not a behaviour to live with.

| | Interface contract | base/base today |
|---|---|---|
| **D1** | `createPolicy(admin, UNION)` → `IncompatiblePolicyType` | `Panic(0x21)` |
| **D2** | `createPolicyWithAccounts(admin, UNION, accts)` → `IncompatiblePolicyType` | `Panic(0x21)` |
| **D3** | `createPolicy(0, UNION)` → `ZeroAddress` outranks the type guard | `Panic(0x21)` — type guard runs first |

**Cause, in one function** (`policy/logic/v2.rs:188-196`): `validate_create_policy_inputs` consults `PolicyType::is_valid()` — hand-narrowed to `BLOCKLIST|ALLOWLIST` in `policy/abi.rs:99-101` — and does so *before* the zero-admin guard. `Panic(0x21)` is the wrong answer on its own terms: `UNION`/`INTERSECT` **are** valid `PolicyType` discriminants, so the enum conversion succeeds; nothing is out of range. `IncompatiblePolicyType` is both what the interface documents and the idiom already used three times elsewhere in that same file.

**Not regressions.** Before composites, `PolicyType` had only `BLOCKLIST|ALLOWLIST`, so a composite discriminant was out of enum range and also produced `Panic(0x21)`. The wire behaviour is unchanged across the feature; what changed is that the interface now *documents* `IncompatiblePolicyType`. Once base/base is reconciled these become ordinary regression tests that must stay green.

**They fail loudly, and none can hide another.** `die()` raises `SystemExit`, so a plain sequence would abort at D1 and D2/D3 would never execute. `_conformance` uses the collect-all pattern from `precompile_invariants.run`: each detector runs and reports independently, then the run exits non-zero with the full list.

Reading the output: `expect_revert` resolves names from the interface's error set and `Panic(uint256)` isn't in it, so these read `got=None want=IncompatiblePolicyType` — the signature of `Panic(0x21)`, not of a call that failed to revert.

## Harness changes

- `create_composite_policy()` — mirrors `create_policy`; composites emit the canonical `PolicyCreated`, verified rather than assumed.
- `composite_children_from(receipt)` — **`assert_events_emitted` is presence-only.** It gathers `topic0` across every recorded tx and never inspects the payload, so a `CompositePolicyUpdated` carrying the wrong child array would pass it. Child-set assertions decode from a specific receipt instead.
- `NonPayable` for `createCompositePolicy` / `updateComposite` in both `_payable_rejected` and `_value_forwarding_rejected`. `error_name` is **pinned** for the new cases: `expect_raw_revert` treats any non-`ContractLogicError` exception as a pass, so unpinned, a node-level rejection would masquerade as `NonPayable`. No composite-support gating needed — the value guard is the first check in dispatch, ahead of version resolution and selector matching.

## Not verified

- **The journey has never been executed** — no RPC. Verified by encoding every function, event and error binding against the artifacts in `out/`, which catches ABI mistakes but not runtime behavior.
- D1/D2/D3 are code-reading against `feat/implement-compound-policy`, not live repros.
- Assumes the smoke deployment is past Cobalt.

Generated with Claude Code

### DIFF
--- a/script/smoke/chain.py
+++ b/script/smoke/chain.py
@@ -560,11 +560,40 @@ class Chain:
         receipt = self.send(self.policy.functions.createPolicyWithAccounts(admin, ptype, accounts), self.deployer)
         return self._policy_id_from(receipt)
 
+    def create_composite_policy(self, admin: ChecksumAddress, ptype: int, child_ids: list[int]) -> int:
+        """Create a UNION/INTERSECT composite over existing simple policies; return the new id.
+
+        Composites are minted through the same `_create` path as simple policies and emit the
+        canonical `PolicyCreated(policyId, creator, policyType)` (verified against
+        `IPolicyRegistry`/`MockPolicyRegistry`), so `_policy_id_from` needs no composite-specific
+        branch. They additionally emit `CompositePolicyUpdated` carrying the full child set —
+        see `composite_children_from` to read it back off the receipt.
+        """
+        receipt = self.send(self.policy.functions.createCompositePolicy(admin, ptype, child_ids), self.deployer)
+        return self._policy_id_from(receipt)
+
     def _policy_id_from(self, receipt: TxReceipt) -> int:
         events = self.policy.events.PolicyCreated().process_receipt(receipt, errors=DISCARD)
         if not events:
             die("PolicyCreated event not found in receipt")
         return int(events[0]["args"]["policyId"])
+
+    def composite_children_from(self, receipt: TxReceipt, policy_id: int | None = None) -> list[int]:
+        """Decode `CompositePolicyUpdated` from a SPECIFIC receipt and return its `childPolicyIds`.
+
+        `assert_events_emitted` is presence-only (topic0 seen somewhere across the run), which
+        cannot distinguish "the child set was replaced with exactly X" from "some composite event
+        fired". This decodes the payload of one tx so a journey can assert the exact array.
+        Pass `policy_id` to select the event for a specific composite when a receipt carries
+        several; otherwise the first (only) one is returned.
+        """
+        events = self.policy.events.CompositePolicyUpdated().process_receipt(receipt, errors=DISCARD)
+        if policy_id is not None:
+            events = [e for e in events if int(e["args"]["policyId"]) == policy_id]
+        if not events:
+            suffix = f" for policyId={policy_id}" if policy_id is not None else ""
+            die(f"CompositePolicyUpdated event not found in receipt{suffix}")
+        return [int(child) for child in events[0]["args"]["childPolicyIds"]]
 
 
 def _norm(v: object) -> object:
--- a/script/smoke/chain.py
+++ b/script/smoke/chain.py
@@ -18,6 +18,7 @@ import sys
 import time
 import urllib.error
 import urllib.request
+from collections.abc import Callable
 
 from eth_account import Account
 from eth_account.signers.local import LocalAccount
@@ -38,6 +39,46 @@ from .provider import ConsistentHTTPProvider
 
 def log(msg: str) -> None:
     print(f"[smoke] {msg}", file=sys.stderr)
+
+
+def run_collected(
+    checks: "list[tuple[str, Callable[[], None]]]",
+    *,
+    label: str,
+    informational: "frozenset[str] | set[str]" = frozenset(),
+    step_prefix: str = "",
+) -> None:
+    """Run every check, recording failures instead of aborting at the first.
+
+    `die()` raises SystemExit, so a plain sequence stops at the first failure and the remaining
+    checks never execute. Each `(name, thunk)` is isolated and its failure recorded. Names in
+    `informational` are reported but do not fail the run. Raises SystemExit at the end if any
+    required check failed.
+    """
+    findings: list[tuple[str, str, bool]] = []
+    for i, (name, thunk) in enumerate(checks, 1):
+        step(f"{step_prefix}{i}", name)
+        required = name not in informational
+        try:
+            thunk()
+        except SystemExit as exc:
+            detail = str(exc.code or "").replace("[smoke] ERROR: ", "")
+            print(f"  ✗ {'FINDING' if required else 'info'}: {detail}", file=sys.stderr)
+            findings.append((name, detail, required))
+        except Exception as exc:  # noqa: BLE001 - harness/RPC error, surface as a finding
+            detail = f"{type(exc).__name__}: {exc}"
+            print(f"  ✗ ERROR: {detail}", file=sys.stderr)
+            findings.append((name, detail, required))
+
+    required_fail = [f for f in findings if f[2]]
+    info_only = [f for f in findings if not f[2]]
+    log(f"{label}: {len(checks) - len(findings)}/{len(checks)} checks held")
+    for name, detail, _ in findings:
+        log(f"  ✗ {name} — {detail}")
+    if info_only:
+        log(f"({len(info_only)} accepted divergence(s) reported as informational)")
+    if required_fail:
+        raise SystemExit(f"[smoke] {label}: {len(required_fail)} finding(s) need triage")
 
 
 def step(n: object, desc: str) -> None:
--- a/script/smoke/chain.py
+++ b/script/smoke/chain.py
@@ -561,14 +561,7 @@ class Chain:
         return self._policy_id_from(receipt)
 
     def create_composite_policy(self, admin: ChecksumAddress, ptype: int, child_ids: list[int]) -> int:
-        """Create a UNION/INTERSECT composite over existing simple policies; return the new id.
-
-        Composites are minted through the same `_create` path as simple policies and emit the
-        canonical `PolicyCreated(policyId, creator, policyType)` (verified against
-        `IPolicyRegistry`/`MockPolicyRegistry`), so `_policy_id_from` needs no composite-specific
-        branch. They additionally emit `CompositePolicyUpdated` carrying the full child set —
-        see `composite_children_from` to read it back off the receipt.
-        """
+        """Create a UNION/INTERSECT composite over existing simple policies; return the new id."""
         receipt = self.send(self.policy.functions.createCompositePolicy(admin, ptype, child_ids), self.deployer)
         return self._policy_id_from(receipt)
 
@@ -579,14 +572,7 @@ class Chain:
         return int(events[0]["args"]["policyId"])
 
     def composite_children_from(self, receipt: TxReceipt, policy_id: int | None = None) -> list[int]:
-        """Decode `CompositePolicyUpdated` from a SPECIFIC receipt and return its `childPolicyIds`.
-
-        `assert_events_emitted` is presence-only (topic0 seen somewhere across the run), which
-        cannot distinguish "the child set was replaced with exactly X" from "some composite event
-        fired". This decodes the payload of one tx so a journey can assert the exact array.
-        Pass `policy_id` to select the event for a specific composite when a receipt carries
-        several; otherwise the first (only) one is returned.
-        """
+        """Decode `CompositePolicyUpdated` from a SPECIFIC receipt and return its `childPolicyIds`."""
         events = self.policy.events.CompositePolicyUpdated().process_receipt(receipt, errors=DISCARD)
         if policy_id is not None:
             events = [e for e in events if int(e["args"]["policyId"]) == policy_id]
--- a/script/smoke/journeys/policy_registry.py
+++ b/script/smoke/journeys/policy_registry.py
@@ -8,16 +8,16 @@ the token-side write-time validation, then a flow-level event check.
 `_composite` covers the UNION/INTERSECT composite gates: construction, live
 gate evaluation over child membership, full-replacement `updateComposite`
 semantics, the child-set validation reverts, and token enforcement through a
-composite. `_known_divergences` runs LAST and is expected to fail against the
+composite. `_regressions` runs LAST and is expected to fail against the
 current live deployment (see the function docstring).
 """
 
 from __future__ import annotations
 
-import sys
+from functools import partial
 
 from .. import config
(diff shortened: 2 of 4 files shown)


script/smoke/journeys/policy_registry.py
```diff
@@ -100,25 +106,197 @@ def _edges(c: Chain, tok, pid_r: int, pid_b: int) -> None:
...
         "PolicyAdminUpdated(uint64,address,address)",
+        "CompositePolicyUpdated(uint64,address,uint64[])",
         "B20Created(address,uint8,string,string,uint8,bytes)",
         "PolicyUpdated(bytes32,uint64,uint64)",
         "Transfer(address,address,uint256)",
         "RoleGranted(bytes32,address,address)",
     )
 
 
+def _known_divergences(c: Chain) -> None:
```
rayyan224: This should be renamed regressions

script/smoke/chain.py
```diff
@@ -560,12 +560,41 @@ def create_policy_with_accounts(self, admin: ChecksumAddress, ptype: int, accoun
         receipt = self.send(self.policy.functions.createPolicyWithAccounts(admin, ptype, accounts), self.deployer)
         return self._policy_id_from(receipt)
 
+    def create_composite_policy(self, admin: ChecksumAddress, ptype: int, child_ids: list[int]) -> int:
+        """Create a UNION/INTERSECT composite over existing simple policies; return the new id.
+
+        Composites are minted through the same `_create` path as simple policies and emit the
```
rayyan224: Just keep Create a UNION/INTERSECT composite over existing simple policies; return the new id.

script/smoke/chain.py
```diff
@@ -560,12 +560,41 @@ def create_policy_with_accounts(self, admin: ChecksumAddress, ptype: int, accoun
...
+        return self._policy_id_from(receipt)
+
     def _policy_id_from(self, receipt: TxReceipt) -> int:
         events = self.policy.events.PolicyCreated().process_receipt(receipt, errors=DISCARD)
         if not events:
             die("PolicyCreated event not found in receipt")
         return int(events[0]["args"]["policyId"])
 
+    def composite_children_from(self, receipt: TxReceipt, policy_id: int | None = None) -> list[int]:
+        """Decode `CompositePolicyUpdated` from a SPECIFIC receipt and return its `childPolicyIds`.
```
rayyan224: Just keep Decode `CompositePolicyUpdated` from a SPECIFIC receipt and return its `childPolicyIds`.

script/smoke/chain.py
rayyan224: Done in e8337ce — trimmed to the summary line.

script/smoke/chain.py
rayyan224: Done in e8337ce — trimmed to the summary line.

script/smoke/journeys/policy_registry.py
```diff
@@ -100,25 +108,314 @@ def _edges(c: Chain, tok, pid_r: int, pid_b: int) -> None:
     c.expect_revert("PolicyNotFound", tok.functions.updatePolicy(config.TRANSFER_SENDER_POLICY, 999999), c.DEPLOYER)
 
 
+def _composite(c: Chain, pid_b: int) -> None:
+    """Composite (UNION / INTERSECT) policies: construction, evaluation, update, edges, enforcement.
+
+    Spec IDs are from brains/composite_policy/SMOKE_TEST_SPECS.md.
```
stevieraykatz: brains?

script/smoke/journeys/policy_registry.py
```diff
@@ -100,25 +108,314 @@ def _edges(c: Chain, tok, pid_r: int, pid_b: int) -> None:
...
+
+    Diagnosing the output: `expect_revert` resolves names from the interface ABI's error set and
+    `Panic(uint256)` is not in it, so these read `got=None want=IncompatiblePolicyType`. That is the
+    signature of `Panic(0x21)` — not of a call that failed to revert.
+
+    Runs LAST so the rest of the journey reports first. All are `eth_call` simulations; none mutates
+    chain state.
+    """
+    findings: list[tuple[str, str]] = []
+    for i, (name, fn) in enumerate(CONFORMANCE_DETECTORS, 33):
```
stevieraykatz: avoid magic number here

script/smoke/journeys/policy_registry.py
```diff
@@ -100,25 +108,314 @@ def _edges(c: Chain, tok, pid_r: int, pid_b: int) -> None:
...
+# fix, not a behaviour to live with. They are listed rather than inlined only so one failure cannot
+# hide the others.
+CONFORMANCE_DETECTORS = [
+    ("D1 createPolicy(admin, UNION) -> IncompatiblePolicyType", _d1_create_policy),
+    ("D2 createPolicyWithAccounts(admin, UNION, accts) -> IncompatiblePolicyType", _d2_create_policy_with_accounts),
+    ("D3 createPolicy(0, UNION) -> ZeroAddress outranks the type guard", _d3_zero_admin_precedence),
+]
+
+
+def _conformance(c: Chain) -> None:
```
stevieraykatz: this whole loop mirrors `run_collected` from `chain.py` but also changes how we do error handling. should import and use the shared helper.

script/smoke/journeys/policy_registry.py
```diff
@@ -100,25 +108,314 @@ def _edges(c: Chain, tok, pid_r: int, pid_b: int) -> None:
...
+        c.DEPLOYER,
+    )
+    # Ghost in FIRST position too: an impl that only validates childPolicyIds[0] passes the case above.
+    c.expect_revert(
+        "PolicyNotFound",
+        c.policy.functions.createCompositePolicy(c.DEPLOYER, config.POLICY_TYPE_UNION, [ghost, pid_x]),
+        c.DEPLOYER,
+    )
+    c.expect_revert("PolicyNotFound", c.policy.functions.updateComposite(pid_union, [pid_x, ghost]), c.DEPLOYER)
+    # CX-8 precedence: existence is checked across the WHOLE set before validity. The invalid child is
```
stevieraykatz: what is CX-8?

script/smoke/journeys/policy_registry.py
```diff
@@ -100,25 +108,314 @@ def _edges(c: Chain, tok, pid_r: int, pid_b: int) -> None:
...
+    # ignored everywhere".
+    c.expect_revert("IncompatiblePolicyType", c.policy.functions.updateComposite(pid_x, [pid_x, pid_y]), c.DEPLOYER)
+    c.expect_revert(
+        "IncompatiblePolicyType", c.policy.functions.updateAllowlist(pid_union, True, [c.BOB]), c.DEPLOYER
+    )
+    c.expect_revert(
+        "IncompatiblePolicyType", c.policy.functions.updateBlocklist(pid_inter, True, [c.BOB]), c.DEPLOYER
+    )
+
+    step(27, "[CX-5/CU-8/CX-8] non-existent child -> PolicyNotFound, and it outranks InvalidChildPolicy")
```
stevieraykatz: ticket/doc refs?

rayyan224: Done in 89387c2 — dropped all 15 spec-ID tags and the `brains/` doc pointer. That doc lives outside this repo, so none of those refs were resolvable from base-std; the step descriptions now stand on their own.

script/smoke/journeys/policy_registry.py
rayyan224: Done in 89387c2 — dropped all 15 spec-ID tags and the `brains/` doc pointer. That doc lives outside this repo, so none of those refs were resolvable from base-std; the step descriptions now stand on their own.

script/smoke/journeys/policy_registry.py
rayyan224: Done in 89387c2 — dropped all 15 spec-ID tags and the `brains/` doc pointer. That doc lives outside this repo, so none of those refs were resolvable from base-std; the step descriptions now stand on their own.

script/smoke/journeys/policy_registry.py
rayyan224: Done in e28e498 — renamed `_conformance` -> `_regressions` and `CONFORMANCE_DETECTORS` -> `REGRESSION_CHECKS`.

script/smoke/journeys/policy_registry.py
rayyan224: Done in e28e498 — `run_collected` now lives in `chain.py` and **both** journeys use it. `precompile_invariants` had the same loop inline with slightly different error handling; that copy and its `_detail` helper are deleted, so the two lanes report identically. Helper takes `(name, thunk)` pairs, honours `informational`, and raises once at the end with every finding.

script/smoke/journeys/policy_registry.py
rayyan224: Done in e28e498 — dropped the hardcoded `33`. The checks now number `R1..R3` via a `step_prefix`, so they don't depend on how many steps precede them. That coupling is what made the earlier "steps 31-32" comment go stale when I inserted steps above it.

stephancill (commented):
lgtm, would like to see proof that these tests are passing before we merge

script/smoke/journeys/policy_registry.py
```diff
@@ -100,25 +108,277 @@ def _edges(c: Chain, tok, pid_r: int, pid_b: int) -> None:
...
+    c.expect_revert(
+        "ZeroAddress",
+        c.policy.functions.createPolicyWithAccounts(config.ZERO, config.POLICY_TYPE_UNION, [c.ALICE]),
+        c.DEPLOYER,
+    )
+
+
+# Checks where base/base does not yet match the `IPolicyRegistry` contract that base-std defines.
+# There are no accepted divergences between the two — every entry is a base/base bug to fix, not a
+# behaviour to live with. Listed rather than inlined so one failure cannot hide the others.
```
stevieraykatz: seems weird to call this out. ok to land with tests that are expected to fail at this point but we shouldn't encode that in the comments.

rayyan224: Removed them entirely in the latest push, rather than just softening the comments. A check that is expected to fail isn't a useful signal in a journey — the fix belongs in base/base. The journey now runs steps 1-32 and is expected to pass end to end.

`chain.run_collected` stays, since `precompile_invariants` uses it — that was the point of extracting it from the inline loop there.

stephancill (approved):

script/smoke/journeys/policy_registry.py
```diff
@@ -100,14 +105,203 @@ def _edges(c: Chain, tok, pid_r: int, pid_b: int) -> None:
...
+    # A composite must be UNION or INTERSECT; a simple gate is rejected by the composite creator.
+    for simple in (config.POLICY_TYPE_ALLOWLIST, config.POLICY_TYPE_BLOCKLIST):
+        c.expect_revert(
+            "IncompatiblePolicyType",
+            c.policy.functions.createCompositePolicy(c.DEPLOYER, simple, [pid_x, pid_y]),
+            c.DEPLOYER,
+        )
+    # Mirror: updateComposite must reject a SIMPLE target, and the membership mutators must reject a
+    # COMPOSITE target. Together with the conformance checks at the end of the journey this closes the
+    # type-guard matrix in both directions, so a red conformance check reads as "the two creator-side
```
robriks, line 221: are these conformance checks still in place? I'm not sure where to see them

script/smoke/journeys/policy_registry.py
```diff
@@ -100,14 +105,203 @@ def _edges(c: Chain, tok, pid_r: int, pid_b: int) -> None:
     c.expect_revert("PolicyNotFound", tok.functions.updatePolicy(config.TRANSFER_SENDER_POLICY, 999999), c.DEPLOYER)
 
 
+def _composite(c: Chain, pid_b: int) -> None:
```
robriks, line 108: should this run a fork/version check ?

robriks (commented):
looking good, some small Qs and +1 on ensuring the LIVE run succeeded

script/smoke/journeys/policy_registry.py
rayyan224: Smokes would always be on the lattest right, so this would be there ?

PR #180, opened 2026-07-28
chore(test): scrub internal ticket/audit IDs from test comments

stevieraykatz:
## Summary

Removes internal identifiers from OSS test comments so nothing internal ships in the public repo, and adds an `AGENTS.md` boundary so it stays that way. Comments/docs only — **no test logic or assertions change**.

- **Linear ticket IDs (`BOP-###`)** stripped from regression/unit test comments. Trailing `Regression: BOP-XXX.` tags removed; temporal references reworded (e.g. "before BOP-230 the privileged path…" → "previously the privileged path…") so the technical intent is preserved.
- **Audit-finding IDs (`L-04`)** removed from `createToken` / `isActivated` test comments the same way.
- Updated the `B20RenamesTest` contract-level doc that described the now-removed `Regression: BOP-XXX.` tagging convention.
- **New `AGENTS.md` boundary**: forbids internal-only references (private docs/links, ticket/audit IDs, internal codenames) from ever entering this public repo. Described *by class* rather than naming internal hosts (which would itself be a leak).

## Scope

8 test files (`transferFrom.t.sol`, `transferFromWithMemo.t.sol`, `mint.t.sol`, `createToken.t.sol`, `isActivated.t.sol`, `B20Removals.t.sol`, `B20Renames.t.sol`, `MockB20.sol`) plus a one-bullet addition to `AGENTS.md`.

## Not included

The `D9` reference in `src/interfaces/IScaledUIAmount.sol` (points at a private design doc) is removed by a separate in-flight PR that rewrites that interface's NatSpec, to avoid a merge conflict on the same block.

## Testing

`forge build` passes; no behavioral changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -94,6 +94,11 @@ Deeper reading: `LIVE_PRECOMPILE_TESTING.md` (cross-validation architecture), `d
 - **Don't reshape mock storage layout unilaterally** — match `MockB20Storage.sol` slot constants
   and cross-validate with `make fork-tests` before merging.
 - **Don't commit `.env` or real private keys** — `.env` is gitignored; use funded testnet keys.
+- **Don't introduce internal-only references** — this repo is public. No links or paths to non-public
+  resources (internal wikis, private design docs, dashboards, chat), no internal ticket or
+  audit-finding IDs, no internal codenames or handles — in code, comments, NatSpec, docs, or commit
+  messages. Restate internal rationale inline in public terms instead of linking it; public EIP/ERC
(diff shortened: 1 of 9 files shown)


stephancill (commented):
can we also add a rule to the AGENTS.md to avoid including internal ticket IDs in code comments?

NullSoldier (approved):

robriks (approved):
LGTM! these are mea culpa, I told claude to reference linear tickets everywhere and didn't realize that contradicted a canonicalized convention !

will work the solidity harness into my local workflow

stephancill (approved):

PR #183, opened 2026-07-29
feat(policy): add compositePolicyChildIds view for composite child sets

rayyan224:
## Summary

Composite policies (UNION/INTERSECT) shipped in #174 with no on-chain read path for their child set. The only ways to recover it today are decoding the `CompositePolicyUpdated` event or a raw `eth_getStorageAt` against slot 4 — the smoke harness does the former (`chain.py: composite_children_from`).

This adds the missing getter to `IPolicyRegistry`, implements it in `MockPolicyRegistry`, and covers it with unit tests.

```solidity
function compositePolicyChildIds(uint64 policyId) external view returns (uint64[] memory);
```

## Semantics

Total — it never reverts, matching the contract every existing getter documents explicitly (`policyExists`, `policyAdmin`, `pendingPolicyAdmin` all say "Never reverts" and degrade to a zero value). An empty array is the array-typed analogue of `address(0)`:

| Input | Returns |
|---|---|
| Existing composite | Stored child set, in the order last written |
| Simple policy (ALLOWLIST / BLOCKLIST) | `[]` |
| Built-in sentinel (`ALWAYS_ALLOW` / `ALWAYS_BLOCK`) | `[]` |
| Well-formed but uncreated ID | `[]` |
| Malformed ID (type byte > INTERSECT) | `[]` |

Returning `[]` rather than reverting `IncompatiblePolicyType` on a non-composite is safe because **an empty return is unambiguous in practice**: both write paths reject a set outside `[2, 4]` with `ChildPoliciesOutsideOfRange`, and there is no clear-the-list path, so a composite that exists always holds at least two children. There's a test pinning that invariant so it can't silently regress.

Ordering is preserved verbatim; the registry neither sorts nor de-duplicates, so a `UNION` over `[a, a]` reads back as `[a, a]`.

## On the name

`childPolicyIds` was the first choice, for exact symmetry with the write path and the event field. It doesn't work: a function of that name collides with the `childPolicyIds` **parameter** on `createCompositePolicy` and `updateComposite`, and solc emits warning 8760 ("This declaration has the same name as another declaration") in `MockPolicyRegistry`. Verified against solc 0.8.30 — `compositePolicyChildIds` compiles clean. The only alternative fix was renaming the mock's parameters away from the interface's, which destroys the symmetry that motivated the name in the first place.

It also reads better at the call site given that a non-composite returns empty.

## Reviewer note

In the mock, the `_isWellFormed` guard **must** precede `_isComposite`. That helper routes through `_typeOf`, whose `PolicyType(uint8(...))` conversion panics with `0x21` on a type byte above `INTERSECT`. The two guards look collapsible but aren't. The Rust precompile compares raw type bytes, so it's safe without the guard — this is a Solidity-side hazard only. Caught by `test_compositePolicyChildIds_success_emptyForMalformedId`, which failed before the guard was added.

## Testing

`forge test` — 136 passed, 0 failed, 4 skipped. `forge fmt --check` clean. 10 new tests in `test/unit/PolicyRegistry/compositePolicyChildIds.t.sol` covering creation-order readback, `updateComposite` full replacement, event/view agreement, duplicate preservation, survival across `renounceAdmin`, the four empty cases (fuzzed for uncreated and malformed IDs), and the never-empty invariant.

## Cross-repo

base-std is the spec forerunner here, so this lands first and `base/base` catches up. The advisory fork-test job (`continue-on-error: true`) is expected to flag `compositePolicyChildIds` as missing until the `base/base` precompile PR merges — the Rust implementation is written and green locally, and I'll link it here once it's open.

Generated with Claude Code

### DIFF
--- a/src/interfaces/IPolicyRegistry.sol
+++ b/src/interfaces/IPolicyRegistry.sol
@@ -260,9 +260,9 @@ interface IPolicyRegistry {
     /// @return Pending admin, or `address(0)`.
     function pendingPolicyAdmin(uint64 policyId) external view returns (address);
 
-    /// @notice Returns the child-policy set of the composite `policyId`, in the order it was last
-    ///         written. Returns an empty array for simple policies, built-in sentinels, unknown
-    ///         IDs, and malformed IDs. Never reverts.
+    /// @notice Returns the child-policy set of the composite `policyId`.
+    ///
+    /// @dev Child-policies are listed in the order they were last written.
     ///
     /// @dev An empty return unambiguously means "not a composite": both write paths reject a set
     ///      outside `[2, 4]` with `ChildPoliciesOutsideOfRange`, so a composite that exists always
--- a/src/interfaces/IPolicyRegistry.sol
+++ b/src/interfaces/IPolicyRegistry.sol
@@ -259,4 +259,19 @@ interface IPolicyRegistry {
     ///
     /// @return Pending admin, or `address(0)`.
     function pendingPolicyAdmin(uint64 policyId) external view returns (address);
+
+    /// @notice Returns the child-policy set of the composite `policyId`, in the order it was last
+    ///         written. Returns an empty array for simple policies, built-in sentinels, unknown
+    ///         IDs, and malformed IDs. Never reverts.
+    ///
+    /// @dev An empty return unambiguously means "not a composite": both write paths reject a set
+    ///      outside `[2, 4]` with `ChildPoliciesOutsideOfRange`, so a composite that exists always
+    ///      holds at least 2 children and there is no clear-the-list path.
+    /// @dev The registry preserves the caller's ordering verbatim and neither sorts nor
+    ///      de-duplicates.
+    ///
+    /// @param policyId Policy to query.
+    ///
+    /// @return Child policy IDs, or an empty array.
+    function compositePolicyChildIds(uint64 policyId) external view returns (uint64[] memory);
 }
--- a/src/interfaces/IPolicyRegistry.sol
+++ b/src/interfaces/IPolicyRegistry.sol
@@ -263,7 +263,8 @@ interface IPolicyRegistry {
     /// @notice Returns the child-policy set of the composite `policyId`.
     ///
     /// @dev Child-policies are listed in the order they were last written.
-    ///
+    /// @dev Returns an empty array for simple policies, built-in sentinels, unknown IDs, and
+    ///      malformed IDs. Never reverts.
     /// @dev An empty return unambiguously means "not a composite".
     /// @dev The registry preserves the caller's ordering verbatim and neither sorts nor
     ///      de-duplicates.
--- a/src/interfaces/IPolicyRegistry.sol
+++ b/src/interfaces/IPolicyRegistry.sol
@@ -264,9 +264,7 @@ interface IPolicyRegistry {
     ///
     /// @dev Child-policies are listed in the order they were last written.
     ///
-    /// @dev An empty return unambiguously means "not a composite": both write paths reject a set
-    ///      outside `[2, 4]` with `ChildPoliciesOutsideOfRange`, so a composite that exists always
-    ///      holds at least 2 children and there is no clear-the-list path.
+    /// @dev An empty return unambiguously means "not a composite".
     /// @dev The registry preserves the caller's ordering verbatim and neither sorts nor
     ///      de-duplicates.
     ///
--- a/test/unit/PolicyRegistry/compositePolicyChildIds.t.sol
+++ b/test/unit/PolicyRegistry/compositePolicyChildIds.t.sol
@@ -1,0 +1,137 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {Vm} from "forge-std/Vm.sol";
+
+import {IPolicyRegistry} from "base-std/interfaces/IPolicyRegistry.sol";
+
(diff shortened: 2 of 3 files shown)


src/interfaces/IPolicyRegistry.sol
```diff
@@ -259,4 +259,19 @@ interface IPolicyRegistry {
     ///
     /// @return Pending admin, or `address(0)`.
     function pendingPolicyAdmin(uint64 policyId) external view returns (address);
+
+    /// @notice Returns the child-policy set of the composite `policyId`, in the order it was last
+    ///         written. Returns an empty array for simple policies, built-in sentinels, unknown
+    ///         IDs, and malformed IDs. Never reverts.
```
stevieraykatz:
```suggestion
    /// @notice Returns the child-policy set of the composite `policyId`.
    ///
    /// @dev Child-policies are listed in the order they were last written.
```

src/interfaces/IPolicyRegistry.sol
```diff
@@ -259,4 +259,19 @@ interface IPolicyRegistry {
...
+    ///
+    /// @dev An empty return unambiguously means "not a composite": both write paths reject a set
+    ///      outside `[2, 4]` with `ChildPoliciesOutsideOfRange`, so a composite that exists always
+    ///      holds at least 2 children and there is no clear-the-list path.
```
stevieraykatz:
```suggestion
    /// @dev An empty return unambiguously means "not a composite".
```

stevieraykatz: Also since we don't duplicate, we're allowing a valid composite policy contructed of two copies of the same allowlist/blocklist. This is probably fine but explicitly allowing it seems a little strange.

test/unit/PolicyRegistry/compositePolicyChildIds.t.sol
```diff
@@ -0,0 +1,137 @@
...
+    /// @dev Malformed-ID short-circuit returns empty, matching the rest of the read surface.
+    function test_compositePolicyChildIds_success_emptyForMalformedId(uint64 seed) public view {
+        uint64 policyId = _malformedPolicyId(seed);
+        assertEq(policyRegistry.compositePolicyChildIds(policyId).length, 0);
+    }
+
+    /// @notice Verifies an existing composite never reports fewer than MIN_CHILD_POLICIES children
+    /// @dev This is what makes an empty return unambiguously mean "not a composite": there is no
+    ///      clear-the-list path, so a created composite can never decay to an empty set.
+    function test_compositePolicyChildIds_success_createdCompositeIsNeverEmpty() public {
```
stevieraykatz: Weird to me that we call this a `success` test when this is the only instance of testing the revert case for `ChildPoliciesOutsideOfRange`.

rayyan224: Agreed

src/interfaces/IPolicyRegistry.sol
rayyan224: Applied both suggestions (thanks for committing them directly) — with one deliberate deviation I want to flag rather than slip in.

The trimmed version dropped **"Never reverts."** entirely. I've restored it as a `@dev`:

```solidity
/// @dev Returns an empty array for simple policies, built-in sentinels, unknown IDs, and
///      malformed IDs. Never reverts.
```

Reasoning: that's a contract guarantee rather than prose, and it's documented explicitly on all three sibling getters — `pendingPolicyAdmin` directly above this one ends with "Never reverts.", as do `policyExists` and `policyAdmin`. Leaving it off would make this the only getter on the read surface that doesn't state it, which reads as a deliberate exception rather than a trim. It's also the load-bearing half of the design: it's *why* a non-composite returns `[]` instead of `IncompatiblePolicyType`.

Happy to cut it if you'd rather the whole read surface get trimmed consistently in a separate pass — just didn't want this one getter to silently diverge.

Minor, take it or leave it: there's now some overlap between `@dev Child-policies are listed in the order they were last written.` and the existing `@dev The registry preserves the caller's ordering verbatim and neither sorts nor de-duplicates.` I left both alone rather than restructure your text — say the word and I'll fold them into one line.

rayyan224: Agreed it's strange, and worth flagging that this PR only *documents* the behavior — it doesn't introduce it. De-duplication was never enforced: `createCompositePolicy` and `updateComposite` have validated child sets since #174 via `_requireCreatedSimplePolicies`, which checks existence and simple-ness per child but never compares children to each other. Same on the Rust side in `logic/v2.rs: validate_composite_child_policies`.

So a `UNION` over `[a, a]` is accepted today. It's semantically harmless — `UNION` and `INTERSECT` are both idempotent, so a duplicate child can't change the authorization result, only waste a slot out of the 4-child budget.

Flagging one interaction: the test trim above **removes `preservesDuplicates`**, which was the only test documenting this. So right now the behavior is neither enforced nor pinned. Three options:

1. **Leave as-is** — undocumented and unenforced. Cheapest, but the next reader has to rediscover it.
2. **Restore the test** as documentation of intentional behavior, without changing semantics.
3. **Reject duplicates** — a real semantics change needing a matching `base` PR (a new error, or reuse `InvalidChildPolicy`), plus a golden re-bless. Since Cobalt hasn't activated anywhere, this is still a free change; after it ships it wouldn't be.

I'd lean (3) if you think a duplicate child is always a caller mistake — it's cheap to enforce now and impossible later — otherwise (2). Happy to take it either way here or file it separately; let me know which and I'll wire it up.

robriks (approved):
good stuff LGTM!

PR #185, opened 2026-07-30
fix(b20-asset): bubble inner-call Panic from `announce` (BOP-477)

robriks:
## Summary

Reconciles a base/base (Rust, authoritative) ↔ base-std (Solidity reference) divergence in `IB20Asset.announce`. When an inner call fails, the Rust precompile propagates **system-level faults raw** (`Panic`/`OutOfGas`/`Fatal`/`SlotOverflow`, via `is_system_error()`) and wraps **only ordinary reverts** as `InternalCallFailed`. The Solidity mock wrapped *every* inner failure — so e.g. `announce([toScaledBalance(overflowing)])` reverts `Panic(0x11)` on-chain but `InternalCallFailed` in the reference.

Chosen direction (Rust is authoritative — OOG must halt to preserve EVM gas semantics, and a Panic is a VM-level fault): **fix base-std to match Rust.**

- `MockB20Asset.announce` re-raises an inner Solidity `Panic(uint256)` (selector `0x4e487b71`) unchanged via `assembly { revert(add(ret, 0x20), mload(ret)) }`; ordinary reverts still wrap as `InternalCallFailed`.
- `IB20Asset.announce` natspec corrected: an inner Panic propagates raw; OOG halts the whole call; any other revert wraps as `InternalCallFailed`. `InternalCallFailed` error doc clarified.
- Added fork conformance test `test_announce_innerPanic_propagatesRaw` — asserts the raw `Panic(0x11)` payload from the **live** precompile (no `livePrecompiles` skip).
- Added a smoke probe (asset journey, step 14b) + reusable `expect_raw_panic` helper asserting the raw `Panic(0x11)` payload via `eth_call`.

## Scope / limitation

- Only the arithmetic-overflow Panic (**0x11**) is reachable as an inner call on **both** sides, so that's what the conformance/smoke assert.
- `OutOfGas`/`Fatal`/`SlotOverflow` are native-precompile faults a Solidity `delegatecall` cannot faithfully reproduce (OOG follows the 63/64 rule and yields empty returndata) — not synthesized in the mock.
- Enum (`0x21`) / array-OOB (`0x32`) have **no reachable inner-call selector**: strict ABI decoding rejects an out-of-range enum as `AbiDecodeFailed` (an ordinary revert) *before* any conversion guard, and nothing produces an array-OOB. (Note: a malformed-enum inner call surfaces a decode-time `Panic(0x21)` in Solidity vs `AbiDecodeFailed` in Rust — not part of the conformance matrix.)

## Test plan

- [x] `forge test --match-contract B20AssetAnnounceTest` (reference / mocks) — 14 pass
- [x] Full `test/unit/B20Asset/*` reference suite — 100 pass; `forge fmt --check` clean
- [x] **Fork conformance** vs a live `anvil --base` (Cobalt) node: `test/unit/B20Asset/announcement/*` — 20 pass, incl. `test_announce_innerPanic_propagatesRaw`
- [x] **Smoke** `asset` journey vs the same live node — green, incl. the step-14b raw `Panic(0x11)` probe

Companion (base/base): a test-only self-pin extension asserting every Panic kind classifies as a system error.

### DIFF
--- a/script/smoke/chain.py
+++ b/script/smoke/chain.py
@@ -36,6 +36,10 @@ from .codec import topic0
 from .errors import ERROR_BY_SELECTOR
 from .provider import ConsistentHTTPProvider
 
+# Solidity built-in `Panic(uint256)` selector. Not a custom error in any ABI, so it is not in
+# `ERROR_BY_SELECTOR`; matched directly by `expect_raw_panic`.
+PANIC_SELECTOR = bytes.fromhex("4e487b71")
+
 
 class Skip(Exception):
     """Raised by a journey to signal it does not apply to this chain (recorded as a skip, not a fail).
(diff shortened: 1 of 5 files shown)


stephancill (commented):
lgtm, would block on approval from @stevieraykatz

stevieraykatz (approved):

PR #186, opened 2026-07-31
refactor(BOP-440): rename seize surface (seizeWithMemo/SEIZE_ROLE/SEIZE_HOLDER_POLICY); drop burnBlocked changes

stephancill:
## Summary

Finalizes the base-std seize surface per the updated BOP-439 decisions, superseding the design merged in #177.

- **Rename** the seize surface across the IB20 interface, the reference impl (`MockB20`), and tests:
  - `transferFromSeizableWithMemo` → `seizeWithMemo`
  - `TRANSFER_FROM_SEIZABLE_ROLE` → `SEIZE_ROLE`
  - `SEIZABLE_ACCOUNT_POLICY` → `SEIZE_HOLDER_POLICY`
  - event `TransferredFromSeizable` → `Seized`
  - (role/policy keccak strings change accordingly)
- **Drop `burnBlockedWithMemo`** entirely (no burn-with-memo variant).
- **Revert `burnBlocked`** to its pre-#177 behavior: blocked check on `TRANSFER_SENDER_POLICY` and the `BURN` pause vector. It is no longer part of the seize operation class; behavior is otherwise unchanged.
- **Remove `burnBlocked` from the `IB20` interface** as a deprecation signal. The reference impl keeps the selector working for back-compat; formal deprecation/removal is announced for the **Denim** hardfork. Retained `burnBlocked` reference tests call it via a concrete `MockB20` cast.

The `SEIZE` pause vector now governs only `seizeWithMemo`. The dedicated seize storage slot (offset 14) is unchanged, so deployed layouts are undisturbed.

## Testing

- `forge build` OK
- `forge test` — 712 passed, 0 failed, 4 skipped
- `forge fmt --check` clean

Seize suites: `seizeWithMemo` (9), `seizeWithMemo_revertOrder` (5). burnBlocked suites reverted to BURN + `TRANSFER_SENDER_POLICY`: `burnBlocked` (7), `burnBlocked_revertOrder` (6).

## Comparison vs pre-seize baseline

Diffed against `main` as of the commit right before the seize work (#177) landed, to confirm the net effect is purely the additive seize surface with `burnBlocked` untouched. Only #177 plus this branch's commits ever touched the core seize files, so the isolation is clean.

- **`burnBlocked` effectively untouched:** implementation body is byte-identical to baseline (same `BURN` pause, `TRANSFER_SENDER_POLICY` check, `AccountNotBlocked`, `BurnedBlocked`). Tests are identical except the call-site cast `token.burnBlocked(...)` → `MockB20(address(token)).burnBlocked(...)`, forced only because `burnBlocked` was removed from the `IB20` type. Only other addition is a deprecation comment.
- **Storage layout safe:** seize slot added at offset 14; the mock-only `initialized` flag moved 14→15. Offsets 0–13 (every real token-state field) are unchanged, so deployed layouts are undisturbed. `initialized` doesn't exist in the Rust precompile, so its relocation is a no-op there. Offset-pin test updated accordingly.
- **Net new surface (additive/intended):** `SEIZE_ROLE`, `SEIZE_HOLDER_POLICY`, `Seized` event, `AccountNotSeizable` error, `seizeWithMemo` + getters, `_requireSeizable`, and the behavior-preserving `_moveBalance` extraction. `burnBlocked` removed from the interface (intended deprecation), impl retained. `burnBlockedWithMemo` nets to zero vs baseline.
- **One intentional cross-cutting change:** `PausableFeature` gains `SEIZE` (index 3), bumping `ALL_FEATURES_PAUSED` 7→15 (pinned by the removals regression). This is the shared/unversioned enum member that trips the frozen-V1 `isPaused(3)` gotcha on the base/base side.

No unexpected changes.


test/lib/mocks/MockB20.sol
```diff
@@ -308,36 +308,35 @@ abstract contract MockB20 is IB20 {
         emit Memo(msg.sender, memo);
     }
 
+    // Deprecated: `burnBlocked` is retained unchanged for back-compat but is no
+    // longer part of the IB20 interface. Its blocked check reads
+    // `TRANSFER_SENDER_POLICY` and it gates on the `BURN` pause vector — it is NOT
+    // part of the seize operation class (that is `seizeWithMemo`). Formal removal
+    // is announced for the Denim hardfork.
```
stephancill: don't need to mention that it is not part of the seize operation class. we should however state that the recommended path is now seize to self and then burn

test/lib/mocks/MockB20.sol
```diff
@@ -308,36 +308,35 @@ abstract contract MockB20 is IB20 {
         emit Memo(msg.sender, memo);
     }
 
+    // Deprecated: `burnBlocked` is retained unchanged for back-compat but is no
+    // longer part of the IB20 interface. Its blocked check reads
+    // `TRANSFER_SENDER_POLICY` and it gates on the `BURN` pause vector. The
+    // recommended path is now `seizeWithMemo` to a treasury/self address followed
+    // by a normal `burn`. Formal removal is announced for the Denim hardfork.
```
stevieraykatz: Do we really want to commit in a comment here what code will land in Denim?

test/lib/mocks/MockB20.sol
```diff
@@ -308,36 +308,35 @@ abstract contract MockB20 is IB20 {
         emit Memo(msg.sender, memo);
     }
 
+    // Deprecated: `burnBlocked` is retained unchanged for back-compat but is no
```
stevieraykatz: Maybe all caps DEPRECATED with a `@notice`?

test/lib/mocks/MockB20.sol
```diff
@@ -308,36 +308,35 @@ abstract contract MockB20 is IB20 {
...
         onlyRole(BURN_BLOCKED_ROLE)
     {
-        _requireSeizable(from);
+        // The point of burnBlocked is to seize from policy-blocked
+        // accounts. Read the transfer-sender policy ID out of the
+        // transfer-side packed slot and reject if the target is
+        // currently authorized. Enforced unconditionally — including
+        // for factory-originated calls during the bootstrap window —
+        // matching the Rust precompile, which carves no `privileged`
+        // exception for this guard.
```
stevieraykatz: This seems like a weird place for this verbosity. We do one-line comments in impls generally unless something is very complicated. This seems more like natspec body content to me.

test/lib/mocks/MockB20Storage.sol
stevieraykatz, line 70: I missed when we moved this into its own storage slot. Curious why?

stephancill: rationale here https://github.com/base/base-std/pull/177#discussion_r3660245896

test/lib/mocks/MockB20.sol
stephancill: removed

robriks (commented):
aligned with these renames

src/interfaces/IB20.sol
```diff
@@ -105,8 +105,13 @@ interface IB20 {
     /// @notice `policyScope` is not a slot this token (or its variant) supports.
     error UnsupportedPolicyType(bytes32 policyScope);
 
-    /// @notice A seize operation (`burnBlocked`, `burnBlockedWithMemo`, or `transferFromSeizableWithMemo`) was
-    ///         called against a `from` that is currently authorized under `SEIZABLE_ACCOUNT_POLICY` (i.e. not blocked).
+    /// @notice `seizeWithMemo` was called against a `from` that is currently authorized under
+    ///         `SEIZE_HOLDER_POLICY` (i.e. not a member of the seize-holder set, so not eligible to be
+    ///         seized from).
```
```suggestion
    ///         `SEIZE_HOLDER_POLICY` (i.e. not a member of the seize-holder set).
```

src/interfaces/IB20.sol
```diff
@@ -207,13 +213,14 @@ interface IB20 {
     /// @return Role constant.
     function BURN_ROLE() external view returns (bytes32);
 
-    /// @notice Required to call `burnBlocked` and `burnBlockedWithMemo`.
+    /// @notice Required to call the deprecated `burnBlocked` (no longer part of this interface; retained for
+    ///         the back-compat impl, with formal removal announced for Denim).
```
stevieraykatz: denim called out here too. must have missed first time through

test/lib/mocks/MockB20.sol
stephancill: AI: Done — converted to a `@notice DEPRECATED` natspec, with the guard rationale moved to `@dev`.

test/lib/mocks/MockB20.sol
stephancill: AI: Moved the rationale out of the impl body into `@dev` natspec; the function body no longer carries the block comment.

src/interfaces/IB20.sol
stephancill: AI: Applied.

src/interfaces/IB20.sol
stephancill: AI: Removed the Denim mention on the `BURN_BLOCKED_ROLE` getter too. Grep-confirmed there are no remaining `Denim` references in the .sol files.

stevieraykatz (approved):

PR #190, opened 2026-08-04
PolicyRegistry: drop ChildPoliciesOutsideOfRange args, add composite child-count getters

rayyan224:
## Summary
- Removes the `(min, max)` args from `IPolicyRegistry.ChildPoliciesOutsideOfRange`; it now reverts with no args.
- Adds two new view functions to `IPolicyRegistry`: `MIN_COMPOSITE_CHILD_POLICIES()` and `MAX_COMPOSITE_CHILD_POLICIES()`, so the valid composite child-count range is discoverable via a call instead of parsed from the error args.
- `MockPolicyRegistry` implements the new getters as `public constant`s aliasing the existing `MIN_CHILD_POLICIES`/`MAX_CHILD_POLICIES` internal constants (same pattern already used for `ALWAYS_ALLOW_ID`/`ALWAYS_BLOCK_ID`), and both revert sites (`createCompositePolicy`, `updateComposite`) now throw the argless error.
- Updates existing `PolicyRegistry` unit tests' revert expectations/NatSpec to match, and adds `compositeChildPolicyLimits.t.sol` covering the two new getters.

## Test plan
- [x] `forge build` — compiles clean, `MockPolicyRegistry` still satisfies `IPolicyRegistry`
- [x] `forge test --match-path "test/unit/PolicyRegistry/*"` — 135 passed, 0 failed
- [x] `forge test` (full suite) — 715 passed, 0 failed

### DIFF
--- a/test/lib/mocks/MockPolicyRegistry.sol
+++ b/test/lib/mocks/MockPolicyRegistry.sol
@@ -80,21 +80,15 @@ contract MockPolicyRegistry is IPolicyRegistry {
     /// @notice Minimum number of child policies a composite must reference.
     ///         `createCompositePolicy` and `updateComposite` revert with
     ///         `ChildPoliciesOutsideOfRange` when `childPolicyIds.length` is below this value.
-    uint256 internal constant MIN_CHILD_POLICIES = 2;
+    uint256 public constant MIN_COMPOSITE_CHILD_POLICIES = 2;
 
     /// @notice Maximum number of child policies a composite may reference.
     ///         `createCompositePolicy` and `updateComposite` revert with
     ///         `ChildPoliciesOutsideOfRange` when `childPolicyIds.length` is outside
-    ///         `[MIN_CHILD_POLICIES, MAX_CHILD_POLICIES]`.
+    ///         `[MIN_COMPOSITE_CHILD_POLICIES, MAX_COMPOSITE_CHILD_POLICIES]`.
     ///         Distinct from `MAX_BATCH_SIZE` (64), which caps account-membership batches.
     ///         Mirrors the Rust PolicyRegistry precompile.
-    uint256 internal constant MAX_CHILD_POLICIES = 4;
-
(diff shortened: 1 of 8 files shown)


test/lib/mocks/MockPolicyRegistry.sol
```diff
@@ -79,18 +79,23 @@ contract MockPolicyRegistry is IPolicyRegistry {
...
-    ///         `ChildPoliciesOutsideOfRange(MIN_CHILD_POLICIES, MAX_CHILD_POLICIES)`
-    ///         when `childPolicyIds.length` is outside `[MIN_CHILD_POLICIES, MAX_CHILD_POLICIES]`.
+    ///         `ChildPoliciesOutsideOfRange` when `childPolicyIds.length` is outside
+    ///         `[MIN_CHILD_POLICIES, MAX_CHILD_POLICIES]`.
     ///         Distinct from `MAX_BATCH_SIZE` (64), which caps account-membership batches.
     ///         Mirrors the Rust PolicyRegistry precompile.
     uint256 internal constant MAX_CHILD_POLICIES = 4;
 
+    /// @inheritdoc IPolicyRegistry
+    uint256 public constant MIN_COMPOSITE_CHILD_POLICIES = MIN_CHILD_POLICIES;
```
rayyan224: Let's remove internal const and use the public one only ?

rayyan224: Done — removed the internal MIN_CHILD_POLICIES/MAX_CHILD_POLICIES and now MIN_COMPOSITE_CHILD_POLICIES/MAX_COMPOSITE_CHILD_POLICIES are the only (public) constants, used directly in the revert checks too. e5a711a

stevieraykatz (approved):

rayyan224:
Tracked in BOP-497: https://linear.app/coinbase/issue/BOP-497/update-naming-for-error-argument-interface-design-review

PR #192, opened 2026-08-04
feat(BOP-495): ERC-8056 interface-review follow-ups (renames + Conversion extension)

robriks:
## Summary

Applies the Aug 4 2026 B20 Interface Review follow-ups to the ERC-8056 scaled-multiplier surface (Solidity interface + reference mock + tests + smoke). Paired in lockstep with base/base PR base/base#4285 — land together.

Scope is **ERC-8056 + multiplier scheduling only**. Every wire change is Cobalt-only (AssetV2, not yet activated on any network) or **add-alias + deprecate** on the frozen Beryl surface — nothing on-chain breaks.

## Changes

- **Rename** `IScaledUIAmount.sol` → `IERC8056.sol` (file only; interface identifiers unchanged).
- **Cobalt-only vocabulary:** `ScheduleOverlap` → `PendingUpdateExists`, `NoScheduledMultiplier` → `NoScheduledUIMultiplier`, `MultiplierUpdateCancelled` → `UIMultiplierUpdateCancelled`.
- **`updateUIMultiplier`** is the canonical instant-failsafe.
- **`IScaledUIAmountConversion`** (`0x57854fc3`) adopted: `toUIAmount` / `fromUIAmount` are the canonical converters, advertised via `supportsInterface`.
- **`MAX_UI_MULTIPLIER()`** getter exposes the `type(uint128).max` setter bound.
- **Dual event on the instant setter:** `updateUIMultiplier` (and the retained `updateMultiplier`) emits **both** the deprecated `MultiplierUpdated(newMultiplier)` and the ERC-8056 `UIMultiplierUpdated`, so indexers on the legacy topic keep working. The scheduled `setUIMultiplier` emits only `UIMultiplierUpdated`.

## Deprecation model (keep in interface, marked deprecated)

Following the team decision (and mirroring base/base-std#193's `burnBlocked` treatment), the legacy methods `updateMultiplier` / `toScaledBalance` / `toRawBalance` are **kept in the `IB20Asset` interface, documented `DEPRECATED.`** — not removed. They remain dialable and aliased under the new names, so block explorers (which need the advertised legacy surface) and developers (who get the canonical names) are both satisfied.

## Test plan
- [x] `forge test` — all pass (4 fork-gated skips)
- [x] `forge fmt --check`, interface-coverage + forge-coverage green
- [x] Base Std Fork Tests (advisory) green
- [ ] Cobalt conformance via base/base#4285 (base_std_ref pinned to this commit)

## Refs
BOP-495 (parent BOP-429 / B20 Improvements). Paired base/base PR: base/base#4285.


robriks:
(fork tests need the names in base/base #4285 to land before passing)

stevieraykatz:
PR Body still describes the previous approach of dropping deprecated methods. Also your local env is leaking into the PR body via `~/work/erc8056...` in the Refs section.

script/smoke/README.md
```diff
@@ -110,9 +110,9 @@ Seven "journeys", run as a whole suite (a single journey can still be run via th
...
-| `multiplier` | ERC-8056 scheduled multiplier (AssetV2 @ Cobalt): `setUIMultiplier` scheduling + its guards (`InvalidMultiplier`, `EffectiveAtInPast`, `EffectiveAtTooFar`, `ScheduleOverlap`), `cancelScheduledMultiplier` (+ `NoScheduledMultiplier`), the `updateMultiplier` instant-failsafe V2 event semantics (`UIMultiplierUpdated` + `MultiplierUpdateCancelled`, *not* `MultiplierUpdated`), the read aliases (`uiMultiplier`/`balanceOfUI`/`totalSupplyUI`), and ERC-165 advertisement. **Skips** cleanly on a pre-Cobalt chain (probed via `supportsInterface(0xa60bf13d)`). |
+| `multiplier` | ERC-8056 scheduled multiplier (AssetV2 @ Cobalt): `setUIMultiplier` scheduling + its guards (`InvalidMultiplier`, `EffectiveAtInPast`, `EffectiveAtTooFar`, `PendingUpdateExists`), `cancelScheduledMultiplier` (+ `NoScheduledUIMultiplier`), the `updateMultiplier` instant-failsafe V2 event semantics (`UIMultiplierUpdated` + `UIMultiplierUpdateCancelled`, *not* `MultiplierUpdated`), the read aliases (`uiMultiplier`/`balanceOfUI`/`totalSupplyUI`), and ERC-165 advertisement. **Skips** cleanly on a pre-Cobalt chain (probed via `supportsInterface(0xa60bf13d)`). |
```
stevieraykatz: this is inaccurate according to the branch in smoke test logic. your test calls `updateUIMultiplier` and asserts `MultiplierUpdated` is still emitted.

robriks: good eye! updated in 399ee93

stevieraykatz (approved):

src/interfaces/IB20Asset.sol
```diff
@@ -22,7 +34,7 @@ interface IB20Asset is IB20, IERC165, IScaledUIAmount, IScaledUIAmountNewUIMulti
     /// @notice `updateExtraMetadata` was called with an empty `key`.
     error InvalidMetadataKey();
 
-    /// @notice A multiplier setter (`setUIMultiplier` or `updateMultiplier`) was called with a
+    /// @notice A multiplier setter (`setUIMultiplier` or `updateUIMultiplier`) was called with a
```
ilikesymmetry: we only have one setter right?
```suggestion
    /// @notice `updateUIMultiplier` was called with a
```

src/interfaces/IB20Asset.sol
```diff
@@ -41,10 +53,10 @@ interface IB20Asset is IB20, IERC165, IScaledUIAmount, IScaledUIAmountNewUIMulti
     /// @notice `setUIMultiplier` was called while a live pending update already exists
     ///
     /// @param pendingEffectiveAt The `effectiveAt` of the live pending update.
-    error ScheduleOverlap(uint256 pendingEffectiveAt);
+    error PendingUpdateExists(uint256 pendingEffectiveAt);
```
ilikesymmetry:
```suggestion
    error UIMultiplierUpdateExists(uint256 effectiveAt);
```

src/interfaces/IB20Asset.sol
```diff
@@ -41,10 +53,10 @@ interface IB20Asset is IB20, IERC165, IScaledUIAmount, IScaledUIAmountNewUIMulti
...
 
     /// @notice `cancelScheduledMultiplier` was called when there is no live pending update
-    error NoScheduledMultiplier();
+    error NoScheduledUIMultiplier();
```
ilikesymmetry: for parity with above error
```suggestion
    error UIMultiplierUpdateDoesNotExist();
```

src/interfaces/IB20Asset.sol
```diff
@@ -209,6 +239,14 @@ interface IB20Asset is IB20, IERC165, IScaledUIAmount, IScaledUIAmountNewUIMulti
     /// @dev Reverts with `InvalidMultiplier` when `newMultiplier` is zero or above `type(uint128).max`.
     ///
     /// @param newMultiplier New multiplier scaled to `WAD_PRECISION`; must be in `(0, type(uint128).max]`.
+    function updateUIMultiplier(uint256 newMultiplier) external;
```
ilikesymmetry: confused, don't we have a `setUIMultiplierUpdate` that we just need to rename? Whatever function we are adding in Cobalt needs to have the `effectiveAt` parameter

src/interfaces/IB20Asset.sol
```diff
@@ -198,7 +228,7 @@ interface IB20Asset is IB20, IERC165, IScaledUIAmount, IScaledUIAmountNewUIMulti
     ///         (`effectiveAt` resets to 0).
     ///
     /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `OPERATOR_ROLE`.
-    /// @dev Reverts with `NoScheduledMultiplier` when there is no live pending update.
+    /// @dev Reverts with `NoScheduledUIMultiplier` when there is no live pending update.
     function cancelScheduledMultiplier() external;
```
ilikesymmetry: I think we also wanted this renamed to include "UI" whenever we mention "Multiplier" and keep consistent with "multiplier update" as the noun that is being cancelled theoretically.
```suggestion
    function cancelUIMultiplierUpdate() external;
```

src/interfaces/IB20Asset.sol
```diff
@@ -41,10 +53,10 @@ interface IB20Asset is IB20, IERC165, IScaledUIAmount, IScaledUIAmountNewUIMulti
     /// @notice `setUIMultiplier` was called while a live pending update already exists
```
ilikesymmetry: something in the prompt seems off to miss all the renames of `setUIMultiplier` to `updateUIMultiplier`

ilikesymmetry (commented):
I'm having second thoughts now on if our setter and cancellation system should be proactively designed for queueing multiple updates. I recall this is a somewhat known use case. If we know we're eventually going to have to build such a thing, I'd prefer to spend an extra cycle to get it right now so we don't have as much to change. Probably worth a longer conversation though.

src/interfaces/IB20Asset.sol
robriks: we (b20 + tequities) had decided to keep the instantaneous `updateMultiplier` function as an admin failsafe, partly because its "tech debt" can't be aliased under the new function

robriks: (as opposed to the new `updateUIMultiplier` that accepts the `effectiveAt` parameter)

src/interfaces/IB20Asset.sol
robriks: there are two, the instantaneous failsafe (legacy) `updateMultiplier` and the newer `updateUIMultiplier`

robriks:
> I'm having second thoughts now on if our setter and cancellation system should be proactively designed for queueing multiple updates. I recall this is a somewhat known use case. If we know we're eventually going to have to build such a thing, I'd prefer to spend an extra cycle to get it right now so we don't have as much to change. Probably worth a longer conversation though.

this was [litigated in a conversation with the tequities team around 3 weeks ago](https://coinbase.slack.com/archives/C09E09VT1NU/p1784054815169649), and the current design leaves room for future extension. 

ofc we can reverse that decision but I toward that being overengineering for a problem that we have not faced yet. it would also bring into question the value of the ongoing audit since a refactor of that size touches more than just the cosmetic naming we're currently refactoring

stephancill (approved):

src/interfaces/IB20Asset.sol
robriks: you're right, that's my bad. I was trying to multitask while working on the ledger product with Rahul and didn't give it my usual oversight. I was thinking this would be a small & simple rename PR, but the refactor scope is actually quite large which led to gaps from the LLM

fixed in [22d67a4](https://github.com/base/base-std/pull/192/commits/22d67a49dfa121c1e63018d3c154da3d37a23625)

src/interfaces/IB20Asset.sol
robriks: good eye! [22d67a4](https://github.com/base/base-std/pull/192/commits/22d67a49dfa121c1e63018d3c154da3d37a23625)

src/interfaces/IB20Asset.sol
robriks: makes sense, [22d67a4](https://github.com/base/base-std/pull/192/commits/22d67a49dfa121c1e63018d3c154da3d37a23625)

PR #194, opened 2026-08-07
docs(BOP-509): document composite policy support in PolicyRegistry

rayyan224:
## Summary
- `docs/PolicyRegistry/README.md` still described `UNION`/`INTERSECT` composition as a "planned for a future hardfork" feature, but `createCompositePolicy`, `updateComposite`, and `compositePolicyChildIds` already ship in `IPolicyRegistry` (see `src/interfaces/IPolicyRegistry.sol`, `test/unit/PolicyRegistry/*Composite*.t.sol`).
- Rewrote the Policy Types section to cover simple vs. composite policies, including the 2–4 child-count bound and the "no composite-of-composite / no built-in child" rules.
- Added `compositePolicyChildIds`, `MIN_COMPOSITE_CHILD_POLICIES`, `MAX_COMPOSITE_CHILD_POLICIES` to the always-callable list, and `createCompositePolicy`/`updateComposite` to the gated-writes list.
- Added "Create Composite Policy" and "Update Composite Children" user-flow sections (sequence diagrams + revert conditions), mirroring the existing simple-policy flows.

Source: [BOP-509](https://linear.app/coinbase/issue/BOP-509/keep-policyregistry-docs-in-sync-with-composite-policies) — flagged by @connerswenberg in Slack as the docs missing composite policy coverage. The ticket also floats a doc/src-diff-detecting harness as a follow-up; scoping that separately rather than bundling it into this docs fix.

## Test plan
- [x] Docs-only change; no code/behavior touched.
- [x] Cross-checked every new claim against `src/interfaces/IPolicyRegistry.sol` and `test/unit/PolicyRegistry/{createCompositePolicy,updateComposite,compositeChildPolicyLimits,compositePolicyChildIds,isAuthorized}.t.sol`.

### DIFF
--- a/docs/PolicyRegistry/README.md
+++ b/docs/PolicyRegistry/README.md
@@ -1,15 +1,22 @@
 # PolicyRegistry
 
-The PolicyRegistry is a singleton precompile for list-based access policies — allowlists and blocklists. Any caller can create a policy and nominate its admin; B20 tokens and other consumers reference policies by `uint64` ID for authorization checks. See [`IPolicyRegistry`](../../src/interfaces/IPolicyRegistry.sol) for the full Solidity interface.
+The PolicyRegistry is a singleton precompile for list-based and composite access policies. Any caller can create a policy and nominate its admin; B20 tokens and other consumers reference policies by `uint64` ID for authorization checks. See [`IPolicyRegistry`](../../src/interfaces/IPolicyRegistry.sol) for the full Solidity interface.
 
 ## Policy Types
 
-Two policy types are supported today:
+Four policy types are supported, split into two kinds:
+
+**Simple** policies decide from an address set:
 
 - **`BLOCKLIST`** — accounts are authorized by default; the admin maintains a list of accounts to explicitly deny.
 - **`ALLOWLIST`** — accounts are denied by default; the admin maintains a list of accounts to explicitly authorize.
 
-Additional types (union / intersect composition of existing policies) are planned for a future hardfork via additive `PolicyType` enum values and sibling creator functions.
+**Composite** policies decide by combining existing simple policies under a logic gate:
+
+- **`UNION`** (OR) — authorized if *any* child policy authorizes the account.
+- **`INTERSECT`** (AND) — authorized only if *every* child policy authorizes the account.
+
+A composite's child set is 2–4 existing simple (`ALLOWLIST`/`BLOCKLIST`) policy IDs — never another composite, and never a built-in sentinel (`ALWAYS_ALLOW`/`ALWAYS_BLOCK`). Composites reference their children live: `isAuthorized` reads current child membership on every call, not a snapshot taken at composite-creation time, so updating a child's membership immediately changes what the composite authorizes.
 
 ## Policy IDs
 
@@ -36,16 +43,21 @@ The `PolicyRegistry` is gated by the [`ActivationRegistry`](../ActivationRegistr
 - `policyExists`
 - `policyAdmin`
 - `pendingPolicyAdmin`
+- `compositePolicyChildIds`
+- `MIN_COMPOSITE_CHILD_POLICIES`
+- `MAX_COMPOSITE_CHILD_POLICIES`
 
 **Gated** — revert with `FeatureNotActivated` while the feature is inactive:
 
 - `createPolicy`
 - `createPolicyWithAccounts`
+- `createCompositePolicy`
 - `stageUpdateAdmin`
 - `finalizeUpdateAdmin`
 - `renounceAdmin`
 - `updateAllowlist`
 - `updateBlocklist`
+- `updateComposite`
 
 Because reads are never gated, a consumer — a B20 token calling `isAuthorized` on transfer, or an indexer reading membership and admin state — sees the same behavior whether or not the feature is active.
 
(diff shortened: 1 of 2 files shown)


docs/PolicyRegistry/README.md
```diff
@@ -1,15 +1,22 @@
...
 - **`BLOCKLIST`** — accounts are authorized by default; the admin maintains a list of accounts to explicitly deny.
 - **`ALLOWLIST`** — accounts are denied by default; the admin maintains a list of accounts to explicitly authorize.
 
-Additional types (union / intersect composition of existing policies) are planned for a future hardfork via additive `PolicyType` enum values and sibling creator functions.
+**Composite** policies decide by combining existing simple policies under a logic gate:
+
+- **`UNION`** (OR) — authorized if *any* child policy authorizes the account.
+- **`INTERSECT`** (AND) — authorized only if *every* child policy authorizes the account.
+
+A composite's child set is 2–4 existing simple (`ALLOWLIST`/`BLOCKLIST`) policy IDs — never another composite, and never a built-in sentinel (`ALWAYS_ALLOW`/`ALWAYS_BLOCK`). Composites reference their children live: `isAuthorized` reads current child membership on every call, not a snapshot taken at composite-creation time, so updating a child's membership immediately changes what the composite authorizes.
```
rayyan224: We can just keep it as "Composites reference their children live: `isAuthorized` reads current child membership on every call. So updating a child's membership immediately changes what the composite authorizes."

docs/PolicyRegistry/README.md
```diff
@@ -88,6 +120,24 @@ sequenceDiagram
...
+sequenceDiagram
+    participant PolicyAdmin
+    participant PolicyRegistry
+
+    PolicyAdmin->>PolicyRegistry: updateComposite(policyId, childPolicyIds)
+    Note over PolicyRegistry: validate children<br>replace child set in full
+    PolicyRegistry-->>PolicyAdmin: emit CompositePolicyUpdated(policyId, updater, childPolicyIds)
+```
+
+`childPolicyIds` is a full replacement, not a merge — a child omitted from the new set no longer governs the composite, even if it governed before. There is no clear-the-list path: the new set must still satisfy the same size and child-validity rules as creation. Because composites read child membership live, `isAuthorized` reflects the new set immediately.
```
rayyan224: Replace with

"`childPolicyIds` is a full replacement, a child omitted from the new set no longer governs the composite.  The new set must still satisfy the same size and child-validity rules as creation"

docs/PolicyRegistry/README.md
rayyan224: Done.

docs/PolicyRegistry/README.md
rayyan224: Done.

docs/PolicyRegistry/README.md
```diff
@@ -1,15 +1,22 @@
...
 - **`BLOCKLIST`** — accounts are authorized by default; the admin maintains a list of accounts to explicitly deny.
 - **`ALLOWLIST`** — accounts are denied by default; the admin maintains a list of accounts to explicitly authorize.
 
-Additional types (union / intersect composition of existing policies) are planned for a future hardfork via additive `PolicyType` enum values and sibling creator functions.
+**Composite** policies decide by combining existing simple policies under a logic gate:
+
+- **`UNION`** (OR) — authorized if *any* child policy authorizes the account.
+- **`INTERSECT`** (AND) — authorized only if *every* child policy authorizes the account.
+
+A composite's child set is 2–4 existing simple (`ALLOWLIST`/`BLOCKLIST`) policy IDs — never another composite, and never a built-in sentinel (`ALWAYS_ALLOW`/`ALWAYS_BLOCK`). Composites reference their children live: `isAuthorized` reads current child membership on every call. So updating a child's membership immediately changes what the composite authorizes.
```
stevieraykatz, line 19: I wasn't aware we were excluding the sentinels (though that makes sense). I think the interface doc is missing this nuance. From `createCompositePolicy` in `IPolicyRegistry` [L124](https://github.com/base/base-std/blob/5a5605a36bc03449bd0d7540353ce842503042a7/src/interfaces/IPolicyRegistry.sol#L124)

```solidity
    /// @dev Child policies must be simple policies (ALLOWLIST or BLOCKLIST), never another composite.
    ///      The child-policy set is capped at 4.
    /// @dev Reverts with `IncompatiblePolicyType` when `policyType` is not UNION or INTERSECT.
    /// @dev Reverts with `ZeroAddress` when `admin` is `address(0)`.
    /// @dev Reverts with `ChildPoliciesOutsideOfRange` when `childPolicyIds.length` is not in
    ///      `[MIN_COMPOSITE_CHILD_POLICIES, MAX_COMPOSITE_CHILD_POLICIES]`.
    /// @dev Reverts with `PolicyNotFound` when any child policy does not exist.
    /// @dev Reverts with `InvalidChildPolicy` when any child policy is not a simple policy or a built-in policy.
    /// @dev Panics with arithmetic overflow (Panic 0x11) when the policy counter has reached its maximum value.
    ///
```

rayyan224: Good catch — the top-level @dev line only called out "never another composite" and left the sentinel exclusion to be inferred from the later revert line. Fixed the NatSpec on both createCompositePolicy and updateComposite to spell out the built-in-sentinel exclusion explicitly, matching the revert conditions and the README.

stevieraykatz (approved):

PR #195, opened 2026-08-07
fix(BOP-510): reject from == to in seizeWithMemo (self-seize)

rayyan224:
## Summary
- `seizeWithMemo`'s balance move is a no-op when `from == to`, but the call still emitted `Transfer`, `Memo`, and `Seized` with no balance change once the address passed both the `SeizeHolder` and `SeizeReceiver` checks — letting a `SEIZE_ROLE` holder pollute the compliance trail with misleading events. No fund loss.
- Rejects `from == to` with `InvalidReceiver` in `MockB20.seizeWithMemo` (shared by the asset and stablecoin mocks), right after the existing zero-receiver check.
- Updates `IB20.seizeWithMemo` natspec with the new revert condition.
- Adds `test_seizeWithMemo_revert_selfSeize` and two revert-order tests (`role_beats_selfSeize`, `selfSeize_beats_blocked`); fixes two existing fuzz tests (`revert_accountNotBlocked`, `revert_receiverPolicyForbids`) that lacked a `from != to` assumption and would occasionally hit the new guard instead of their intended revert.
- Extends the `seize` smoke journey (`script/smoke/journeys/seize.py`) with a self-seize edge case.
- Mirrors the companion fix in [base](https://github.com/base/base) (Rust precompile).

## Linear
[BOP-510](https://linear.app/coinbase/issue/BOP-510/self-seize-from-to-produces-misleading-events) — Cantina finding L-04.

## Test plan
- [x] `forge test` — all 724 tests pass (0 failed).
- [x] `forge test --match-contract B20SeizeWithMemo` — 22/22 pass, including the new self-seize and revert-order tests.
- [x] `forge fmt --check` clean on all changed files.

### DIFF
--- a/src/interfaces/IB20.sol
+++ b/src/interfaces/IB20.sol
@@ -462,7 +462,8 @@ interface IB20 {
     ///      unconfigured token may seize to any destination (a treasury need not be allowlisted).
     /// @dev Reverts with `ContractPaused(SEIZE)` when `SEIZE` is paused.
     /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `SEIZE_ROLE`.
-    /// @dev Reverts with `InvalidReceiver` when `to == address(0)`.
+    /// @dev Reverts with `InvalidReceiver` when `to == address(0)` or `from == to` (a self-seize is a
+    ///      no-op move that would otherwise still emit a misleading `Transfer`/`Memo`/`Seized`).
     /// @dev Reverts with `AccountNotSeizable` when `from` is currently authorized under `SEIZE_HOLDER_POLICY`.
     /// @dev Reverts with `PolicyForbids(SEIZE_RECEIVER_POLICY, ...)` when `to` is not authorized under `SEIZE_RECEIVER_POLICY`.
     /// @dev Reverts with `InsufficientBalance` when `from`'s balance is below `amount`.
--- a/src/interfaces/IB20.sol
+++ b/src/interfaces/IB20.sol
@@ -462,8 +462,7 @@ interface IB20 {
     ///      unconfigured token may seize to any destination (a treasury need not be allowlisted).
     /// @dev Reverts with `ContractPaused(SEIZE)` when `SEIZE` is paused.
     /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `SEIZE_ROLE`.
-    /// @dev Reverts with `InvalidReceiver` when `to == address(0)` or `from == to` (a self-seize is a
-    ///      no-op move that would otherwise still emit a misleading `Transfer`/`Memo`/`Seized`).
+    /// @dev Reverts with `InvalidReceiver` when `to == address(0)` or `from == to`.
     /// @dev Reverts with `AccountNotSeizable` when `from` is currently authorized under `SEIZE_HOLDER_POLICY`.
     /// @dev Reverts with `PolicyForbids(SEIZE_RECEIVER_POLICY, ...)` when `to` is not authorized under `SEIZE_RECEIVER_POLICY`.
     /// @dev Reverts with `InsufficientBalance` when `from`'s balance is below `amount`.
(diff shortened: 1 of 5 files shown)


src/interfaces/IB20.sol
```diff
@@ -462,7 +462,8 @@ interface IB20 {
     ///      unconfigured token may seize to any destination (a treasury need not be allowlisted).
     /// @dev Reverts with `ContractPaused(SEIZE)` when `SEIZE` is paused.
     /// @dev Reverts with `AccessControlUnauthorizedAccount` when the caller does not hold `SEIZE_ROLE`.
-    /// @dev Reverts with `InvalidReceiver` when `to == address(0)`.
+    /// @dev Reverts with `InvalidReceiver` when `to == address(0)` or `from == to` (a self-seize is a
+    ///      no-op move that would otherwise still emit a misleading `Transfer`/`Memo`/`Seized`).
```
```suggestion
    /// @dev Reverts with `InvalidReceiver` when `to == address(0)` or `from == to`.
```

In general as soon as a natspec comment goes multi-line question the verbosity. Sometimes its warranted but it's a good scanning heuristic.

test/lib/mocks/MockB20.sol
```diff
@@ -332,14 +332,15 @@ abstract contract MockB20 is IB20 {
...
-    ///      here is explicit.
+    ///      would be a burn), `from != to` (otherwise `_moveBalance` is a no-op that would still emit a
+    ///      misleading `Transfer`/`Memo`/`Seized`), and — unlike a normal transfer — no sender/receiver/
+    ///      executor transfer policy is consulted, no allowance is spent, and `from` is not zero-checked
+    ///      (consistent with the burn-blocked family; a zero/empty `from` fails the seizable or balance
+    ///      check anyway). The membership checks are that `from` is blocked under `SEIZE_HOLDER_POLICY`
+    ///      and `to` is authorized under `SEIZE_RECEIVER_POLICY` (mirroring `MINT_RECEIVER_POLICY`; an
+    ///      unset slot is always-allow, so a treasury need not be allowlisted by default). Deliberately
+    ///      does NOT reuse the factory-bootstrap privileged path (which would silently skip the receiver
+    ///      policy); every skip here is explicit.
```
stevieraykatz: Take a cleanup pass on this block. I know it's mostly legacy and not in scope for this PR but it could be slimmed down to about half the length imo.

test/unit/B20/supply/seizeWithMemo.t.sol
```diff
@@ -53,11 +53,30 @@ contract B20SeizeWithMemoTest is B20Test {
...
+    /// @notice Reverts InvalidReceiver when `from == to`, even when both the seizable and receiver
+    ///         checks would otherwise pass. A self-seize is a no-op balance move that would otherwise
+    ///         still emit a misleading `Transfer`/`Memo`/`Seized`, polluting the compliance trail.
+    function test_seizeWithMemo_revert_selfSeize(address account, uint256 amount) public {
+        _assumeValidActor(account);
+        amount = bound(amount, 0, B20Constants.MAX_SUPPLY_CAP);
+        _mint(account, amount);
+        _armSeize();
+        // `account` is seizable (blocked) and SEIZE_RECEIVER_POLICY is unset (always-allow), so both
+        // membership checks would pass — the self-seize guard must fire before either is consulted.
```
stevieraykatz: Confusing inline honestly... consider removing.

test/lib/mocks/MockB20.sol
rayyan224: High level, i think can be simplifed to: 

Reassigns the balance of a user to a different user. Seized users are those whose SEIZE_HOLDER_POLICY alllowed policy returns true, SEIZE_RECEIVER_POLICY are those who can receive.

rayyan224: Trimmed it down per both suggestions — dropped the factory-bootstrap-path aside and folded the rest into a 3-line @dev.

test/unit/B20/supply/seizeWithMemo.t.sol
rayyan224: Removed — agreed, the @notice above already covers it.

robriks (approved):
good stuff!

surfaced a side question on the rust PR about 0-value seizures, I'll run it by Roger

stevieraykatz (approved):

PR #201, opened 2026-08-12
docs(changelog): add root CHANGELOG.md overview across Cobalt entries

rayyan224:
## Summary
- Adds a root `CHANGELOG.md` — a one-page, per-hardfork overview that ties together the three existing per-feature entries under `changelog/cobalt/` (scheduled multiplier updates, seize surface, composite policies), plus a migration checklist.
- Updates the root `README.md` Changelog section to link the new overview first, then the detailed `changelog/` index for full mapping tables and edge-case Q&A.

Requested by Conner/Sohe: the individual per-feature changelogs are good, but there was no higher-level summary tying them together.

## Test plan
- [x] Links in `CHANGELOG.md` and `README.md` resolve to existing files
- N/A — docs only, no code/behavior change


robriks (approved):
nice, this looks great. 

for my own understanding, the root `CHANGELOG.md` file will be used as a forward-looking cursor pointing at the **upcoming** hard fork's changes? ie when Cobalt goes live, that file will start pointing at upcoming Denim changes?

robriks (commented):
nice, I think this is the right structure we should commit to.

is the ordinal system truly necessary? seems like we dont' need to prepend `02` to the file names (as `02_Cobalt_interface_feature.md`) when `Cobalt` is already in the name

rayyan224:
https://github.com/base/base-std/pull/201#pullrequestreview-4921684030
02 is for ordering of the files

PR #214, opened 2026-09-02
feat(b20): rename SEIZE_HOLDER_POLICY to SEIZE_EXEMPT_POLICY

rayyan224:
## Summary
- Rename the seize-from policy scope from `SEIZE_HOLDER_POLICY` to `SEIZE_EXEMPT_POLICY` so the identifier matches the inverted check: authorized accounts are seize-exempt.
- Update the `IB20` getter, `B20Constants` keccak preimage, mocks, tests, smoke journey, and Cobalt changelog (selector `0xfeb346ec`, value `keccak256("SEIZE_EXEMPT_POLICY")`).
- Guard the rename: `SEIZE_HOLDER_POLICY()` must not resolve; `SEIZE_EXEMPT_POLICY()` must match the new digest.

This is a Cobalt wire change (the scope hash and getter selector both change). The Rust precompile in `base/base` needs the same preimage before live-precompile tests can pass.

## Test plan
- [ ] `forge test` (735 passed locally)
- [ ] `forge fmt --check`
- [ ] Confirm `token.SEIZE_EXEMPT_POLICY() == keccak256("SEIZE_EXEMPT_POLICY")`
- [ ] Confirm `SEIZE_HOLDER_POLICY()` no longer resolves
- [ ] Land the matching constant rename in `base/base` before `base-forge test` / `make fork-tests`


stephancill (approved):

changelog/02_Cobalt_B20_seize.md
```diff
@@ -179,7 +179,7 @@ The field is packed into a single 256-bit slot:
 
 | Bits    | Lane | Field      | Scope                   |
 | ------- | ---- | ---------- | ----------------------- |
-| 0–63    | 0    | `seizable` | `SEIZE_HOLDER_POLICY`   |
+| 0–63    | 0    | `seizable` | `SEIZE_EXEMPT_POLICY`   |
```
ilikesymmetry: Naming doesn't make sense between seizable and policy name

test/lib/mocks/MockB20Storage.sol
```diff
@@ -325,7 +325,7 @@ library MockB20Storage {
         return uint256(senderId) | (uint256(receiverId) << 64) | (uint256(executorId) << 128);
     }
 
-    /// @notice Extracts the seize-holder policy id (lane 0) from the seize packed slot.
+    /// @notice Extracts the seize-exempt policy id (lane 0) from the seize packed slot.
     function seizablePolicyId(uint256 packed) internal pure returns (uint64) {
```
ilikesymmetry: Feels like cruft we should clean up when you have downtime. There are two policies under seize operation

changelog/02_Cobalt_B20_seize.md
rayyan224: good catch, the packed struct filed seems to be not caught. Updated

test/lib/mocks/MockB20Storage.sol
rayyan224: Fixed now, later sometimes never comes hahah

PR #215, opened 2026-09-02
feat(fork): select Base upgrade via BASE_UPGRADE env

rayyan224:
## What

The fork-test harness (`script/fork/__main__.py`) hardcoded a bare `--base` when
launching anvil, so it always registered anvil's **default** upgrade. This adds
an optional `BASE_UPGRADE` env var that, when set, is forwarded as
`--base <upgrade>` (e.g. `--base beryl`).

Unset/empty preserves today's behavior (bare `--base` → default upgrade), so
existing callers are unaffected.

## Why

anvil's `--base` flag can now pin a specific historical upgrade at runtime. That
lets a single current base-anvil revision test the latest precompile
implementation against each frozen historical base-std suite, instead of pinning
a distinct anvil revision per hardfork. This harness change is what actually
selects the upgrade at launch; CI passes the fork to test via `BASE_UPGRADE`.

## Changes

- `script/fork/__main__.py`: read `BASE_UPGRADE`, thread it into
  `anvil_running(...)`, build the `--base [<upgrade>]` argv conditionally, and
  log the selected upgrade.
- `script/fork/README.md`: document `BASE_UPGRADE` in the Environment table.

## How to test

Requires patched anvil/forge built from a base-anvil revision that supports
`--base <upgrade>`:

```bash
BASE_UPGRADE=beryl ANVIL_BIN=<…/anvil> FORGE_BIN=<…/forge> ANVIL_LOG=/tmp/a.log \
  make fork-tests ARGS="--match-test <any>"
```

- `[run-fork-tests] base upgrade: beryl` appears and anvil boots.
- Repeat with `BASE_UPGRADE=cobalt` and with it unset (unset → default, unchanged).
- An invalid value makes anvil exit during startup and the harness reports the
  environment error — confirming the value is actually forwarded.

### DIFF
--- a/script/fork/README.md
+++ b/script/fork/README.md
@@ -54,6 +54,7 @@ PYTHONPATH=script script/smoke/.venv/bin/python -m fork --match-test test_transf
 | `PORT` | `8546` | local RPC port for anvil |
 | `ACTIVATION_ADMIN` | `0x9965…A4dc` | address authorized to activate features |
 | `ANVIL_LOG` | `/tmp/anvil.log` | anvil stdout/stderr log path |
+| `BASE_UPGRADE` | _(none)_ | historical Base upgrade to register (e.g. `beryl`, `cobalt`), forwarded as `--base <upgrade>`; unset uses anvil's default upgrade (bare `--base`) |
 | `SKIP_ACTIVATE` | _(none)_ | comma-separated feature names or `0x` ids to leave un-activated (exercises the inactive-feature dispatch path); matched case-insensitively |
 
 ## Exit codes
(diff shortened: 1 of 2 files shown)


script/fork/README.md
```diff
@@ -54,6 +54,7 @@ PYTHONPATH=script script/smoke/.venv/bin/python -m fork --match-test test_transf
 | `PORT` | `8546` | local RPC port for anvil |
```
dancaobangthichdanhnhau-boop: bombuacbvn@yahoo.com

PR #222, opened 2026-09-10
feat(policy): add NOT/invert policy semantics to reference

rayyan224:
## What

Adds an **invert (NOT)** flag on the high bit of the `uint64` policy ID so one membership set can be evaluated as **include or exclude** without maintaining a mirror list. When the bit (`B20Constants.POLICY_INVERT_BIT`, bit 63) is set, `isAuthorized` resolves the base policy (`id & ~POLICY_INVERT_BIT`) and returns the **opposite** of its decision. The base's members are shared, never copied — updating the base updates the inverse.

This is **Option 2a** (invert bit on the ID) with a base-existence check. No new functions or selectors are added; the invert bit reinterprets the existing ID argument.

## Why fail-closed matters

An inverted ID over an **unknown or malformed base returns `false`**, never allow-everyone. Without this guard, a garbage/typo'd ID with the bit set would authorize every account — a mint/transfer/seize bypass. The flip is only applied after the base resolves against a real policy.

## Changes

- **`src/lib/B20Constants.sol`** — `POLICY_INVERT_BIT` (single source of truth) + a pure `invertPolicy()` helper. Negation is pure bit math, so no registry selector is added (kept off the frozen ABI surface).
- **`src/interfaces/IPolicyRegistry.sol`** — NatSpec only (no signature changes): the ID layout + invert flag, `isAuthorized`'s fail-closed rule, strip-to-base semantics on the read getters (`policyExists(~id) == policyExists(id)`, etc.), and the composite create/update contract (a child may carry the flag for "A AND NOT X"; an inverted **composite** child is rejected to preserve the flat-tree invariant).
- **`test/lib/mocks/MockPolicyRegistry.sol`** — invert handling in `_isAuthorized` (fail-closed flip), strip-to-base in the getters, composite-child validation on the base. Non-inverted paths are byte-identical.
- **`test/unit/PolicyRegistry/isAuthorizedInvert.t.sol`** — new suite, fail-closed invariants first, then simple/built-in truth tables, `INTERSECT[A, ~X]`, child validation, getter strip semantics, and the helper.

## Scope

base-std **reference mock** only; the Rust precompile is unchanged. Per AGENTS.md the mock must mirror the Rust impl slot-for-slot — the Rust side does not implement NOT yet, so live-precompile (`base-forge test`) will diverge for the *new* inverted-ID cases until the precompile lands. Existing behavior stays in parity.

## Testing

- `forge build` clean; `forge fmt --check` clean.
- `forge test` — **751 passed, 0 failed, 4 skipped** (adds 16 invert cases).
- `python3 script/check-coverage.py` — all interface functions covered.

### DIFF
--- a/changelog/03_Denim_PolicyRegistry_not_policy.md
+++ b/changelog/03_Denim_PolicyRegistry_not_policy.md
@@ -7,17 +7,17 @@
 
 ## Summary
 
-Issuers may want the inverse of a specific list without maintaining two lists. For example, they may authorize an account only when it is not on a sanctions blocklist. This change adds a query-time invert (NOT) flag in bit 63 of a `uint64` policy ID. When that bit is set, `isAuthorized` resolves the base policy and returns the opposite of that policy's decision.
+This change allows any policy ID to reference the opposite (NOT) of its original outcome at query time. When bit 63 is set, `isAuthorized` resolves the base policy and returns the opposite of that policy's decision.
 
 Members stay on the base policy and are shared, not copied, so an update to the base updates its inverse. Invert therefore creates no new record, no new create path, and no extra storage load (`SLOAD`). The flag applies to every policy type: `ALLOWLIST`, `BLOCKLIST`, and `UNION` / `INTERSECT` composites. 
 
 ## Motivation
 
-Issuers may want the inverse of a specific list, and the registry cannot express that. They may authorize an account only when it is not on a sanctions blocklist, or only when it is not Know Your Customer (KYC) verified. Composites often need the same negation: "allowed to transfer" is frequently "on list A and not on list B", where list B is a sanctions list, a blocked list, or a non-KYC'd list. There is no way to say "the opposite of this policy."
+Issuers may want the inverse of a specific list, and the registry cannot express that. They may authorize an account only when it is not on a sanctions blocklist, or only when it is not Know Your Customer (KYC) verified. Composites often need the same negation: "allowed to transfer" is frequently "on list A and not on list B", where list B is a sanctions list, a blocked list, or a non-KYC'd list. They may also need both sides inverted: "not on list A and not on list B". There is no way to say "the opposite of this policy."
 
-The workaround is to maintain two mirror lists: an allowlist and a blocklist seeded with the same addresses. Every membership change must land on both lists. Any lag rejects valid accounts or admits invalid ones. Composite policies do not remove that second list.
+Without invert, the only way to get the opposite outcome is to create a second policy of the other type and copy the same addresses into it: an allowlist mirrored as a blocklist, or the reverse. Every membership change must then land on both policies. If one update lags, valid accounts are rejected or invalid ones are admitted. A composite that needs "NOT A" still has to point at that second, mirrored policy. It cannot reuse A.
 
-The goal is to let one membership set be evaluated as include or exclude, so issuers never maintain two policies for the same address group.
+The goal is to let one membership set be evaluated as include or exclude, so issuers never maintain two or more policies for the same address group. A composite can invert one child or several: "A AND NOT B", or "NOT A AND NOT B".
 
 ## Background
 
@@ -44,9 +44,6 @@ The structure of a policy ID is:
 
 - Bits `[0:55]` hold a unique counter value. The type is not stored in a slot.
 - Bits `[56:63]` are reserved for `PolicyType`. Only four types are used today (`0–3`: `BLOCKLIST`, `ALLOWLIST`, `UNION`, `INTERSECT`), occupying bits `56–57`. Bits `58–63` are unused.
-
-Built-in sentinels are `ALWAYS_ALLOW` (id 0) and `ALWAYS_BLOCK` (id 1). The counter starts at 2.
-
 
 ## Specs
 
--- a/changelog/03_Denim_PolicyRegistry_not_policy.md
+++ b/changelog/03_Denim_PolicyRegistry_not_policy.md
@@ -1,0 +1,183 @@
+# NOT / Invert Policies
+
+- **Feature Name**: not_policy
+- **Start Date**: 2026-09-09
+- **Authors**: Rayyan Alam
+- **Title**: NOT / Invert Policies
+
+## Summary
+
+Issuers may want the inverse of a specific list without maintaining two lists. For example, they may authorize an account only when it is not on a sanctions blocklist. This change adds a query-time invert (NOT) flag in bit 63 of a `uint64` policy ID. When that bit is set, `isAuthorized` resolves the base policy and returns the opposite of that policy's decision.
+
+Members stay on the base policy and are shared, not copied, so an update to the base updates its inverse. Invert therefore creates no new record, no new create path, and no extra storage load (`SLOAD`). The flag applies to every policy type: `ALLOWLIST`, `BLOCKLIST`, and `UNION` / `INTERSECT` composites. 
+
+## Motivation
+
+Issuers may want the inverse of a specific list, and the registry cannot express that. They may authorize an account only when it is not on a sanctions blocklist, or only when it is not Know Your Customer (KYC) verified. Composites often need the same negation: "allowed to transfer" is frequently "on list A and not on list B", where list B is a sanctions list, a blocked list, or a non-KYC'd list. There is no way to say "the opposite of this policy."
+
+The workaround is to maintain two mirror lists: an allowlist and a blocklist seeded with the same addresses. Every membership change must land on both lists. Any lag rejects valid accounts or admits invalid ones. Composite policies do not remove that second list.
+
+The goal is to let one membership set be evaluated as include or exclude, so issuers never maintain two policies for the same address group.
+
+## Background
+
+### Policy Registry
+
+The Policy Registry is a singleton precompile at `0x8453000000000000000000000000000000000002`. B20 tokens call it for pre-operation compliance checks on an address.
+
+B20 stores a `uint64` policy ID per scope (`TRANSFER_FROM`, `TRANSFER_TO`, `MINT_RECEIVER`, `SEIZE_EXEMPT`, and other scopes) and calls `isAuthorized(policyId, account)` before gated operations.
+
+`isAuthorized` never reverts. A malformed or unknown ID returns `false` (deny).
+
+### Policy ID layout
+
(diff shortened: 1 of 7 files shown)


src/interfaces/IPolicyRegistry.sol
```diff
@@ -127,6 +127,8 @@ interface IPolicyRegistry {
     /// @dev Child policies must be simple policies (ALLOWLIST or BLOCKLIST), never another composite
     ///      and never a built-in sentinel (ALWAYS_ALLOW / ALWAYS_BLOCK). The child-policy set is
```
rayyan224, line 132: We should have a constant here for the INVERTED_POLICY_BIT

test/lib/mocks/MockPolicyRegistry.sol
```diff
@@ -22,6 +22,10 @@ library PolicyRegistryConstants {
     /// @dev    Encodes as an ALLOWLIST at counter 1 (empty allowlist → block all).
     uint64 internal constant ALWAYS_BLOCK_ID = (uint64(uint8(IPolicyRegistry.PolicyType.ALLOWLIST)) << 56) | 1;
 
+    /// @notice High bit of a `uint64` policy ID that inverts the base policy.
+    /// @dev    All view functions see an inverted ID as an extension of the base — policy ID
+    uint64 internal constant INVERTED_POLICY_BIT = uint64(1) << 63;
```
stephancill, line 27: `INVERTED_POLICY_MASK` might be more accurate

stevieraykatz: I don't think "mask" is correct since we're using it as a switch not as a mask. if we applied the entire bit field to an underlying bitfield then it would be a mask, but in this case, it's acting as a single purpose bool packed into reserved bitspace.

src/interfaces/IPolicyRegistry.sol
```diff
@@ -280,9 +288,21 @@ interface IPolicyRegistry {
...
     function compositePolicyChildIds(uint64 policyId) external view returns (uint64[] memory);
+
+    /// @notice Returns `policyId` with its invert flag (bit 63) flipped. Never reverts
+    ///         and reads no state.
+    ///
+    /// @dev This call does not check that `policyId` exists; a missing
+    ///      or malformed base is denied later, at `isAuthorized`.
+    ///
+    /// @param policyId Policy to invert.
+    ///
```
rayyan224, line 305: Add Nat spec to give the NAT spec that inversion twice results in the same

src/interfaces/IPolicyRegistry.sol
```diff
@@ -5,6 +5,10 @@ pragma solidity >=0.8.20 <0.9.0;
 ///
 /// @notice Singleton registry of simple and composite policies. Policies are referenced by
 ///         `uint64 policyId` and queried via `isAuthorized(policyId, account)`.
+///
+/// @dev Invert (`invertedPolicyId`): all view functions see an inverted policy ID as an
```
rayyan224, line 9: Move those to each view functions.

stephancill: Motivation: feature level natspec should be as close to the feature code as possible

test/lib/mocks/MockPolicyRegistry.sol
```diff
@@ -358,6 +384,12 @@ contract MockPolicyRegistry is IPolicyRegistry {
     ///      `_isAuthorized` per child, each of which resolves via the simple path
     ///      (or a built-in short-circuit).
     function _isAuthorized(uint64 policyId, address account) internal view returns (bool) {
+        bool isInverted = policyId & INVERTED_POLICY_BIT != 0;
```
rayyan224, line 387: Move this to 399

changelog/03_Denim_PolicyRegistry_not_policy.md
```diff
@@ -0,0 +1,180 @@
...
+flowchart TD
+    C["createCompositePolicy / updateComposite child"] --> S["_basePolicyId: clear bit 63"]
+    S --> E{"policyExists(base)?"}
+    E -->|no| NF["revert PolicyNotFound (whole set first)"]
+    E -->|yes| T{"base is ALLOWLIST or BLOCKLIST?"}
+    T -->|no, composite| IC["revert InvalidChildPolicy"]
+    T -->|yes| OK["Accept child ID as stored, invert bit kept"]
+```
+
+#### State / gas
```
rayyan224, line 119: Make heading the same as Interfaces and seperate

src/interfaces/IPolicyRegistry.sol
rayyan224: Documented on `invertedPolicyId`: inversion is involutive, so `invertedPolicyId(invertedPolicyId(id)) == id`.

src/interfaces/IPolicyRegistry.sol
rayyan224: Moved the invert NatSpec off the interface header onto `policyExists`, `policyAdmin`, `pendingPolicyAdmin`, `compositePolicyChildIds`, and `isAuthorized` (already had it).

test/lib/mocks/MockPolicyRegistry.sol
rayyan224: Invert now runs after the `ALWAYS_ALLOW` / `ALWAYS_BLOCK` short-circuits. A negated always-allow ID is a different value, so it no longer risks matching the sentinel constant. `_isWellFormed` still runs on the stripped base via the invert path, because bit 63 would otherwise make every inverted ID look malformed.

changelog/03_Denim_PolicyRegistry_not_policy.md
rayyan224: Promoted to `### State / Gas`, same heading level as `### Interface Changes`, still one combined section.

changelog/03_Denim_PolicyRegistry_not_policy.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Summary
+
+This change allows any policy ID to reference the opposite (NOT) of its original outcome at query time. When bit 63 is set, `isAuthorized` resolves the base policy and returns the opposite of that policy's decision.
+
+Members stay on the base policy and are shared, not copied, so an update to the base updates its inverse. Invert therefore creates no new record, no new create path, and no extra storage load (`SLOAD`). The flag applies to every policy type: `ALLOWLIST`, `BLOCKLIST`, and `UNION` / `INTERSECT` composites. 
+
+## Motivation
+
+Issuers may want the inverse of a specific list, and the registry cannot express that. They may authorize an account only when it is not on a sanctions blocklist, or only when it is not Know Your Customer (KYC) verified. Composites often need the same negation: "allowed to transfer" is frequently "on list A and not on list B", where list B is a sanctions list, a blocked list, or a non-KYC'd list. They may also need both sides inverted: "not on list A and not on list B". There is no way to say "the opposite of this policy."
```
rayyan224: I would love to see this motvation section be changed to highlight more of the fundamental issue we are solving.  Something like below thought ?

Increasingly, we’re seeing the basic Policy Registry used less as a standalone ruleset and more as a shared registry of addresses that other policies compose around.

Inverse Policies build on that model by allowing a single address list to represent either side of a rule. The same underlying list can behave as an allowlist or a blocklist depending on whether the policy is evaluated normally or inverted.

For example, an issuer may maintain a shared KYC list. Whether that list should be interpreted as “only these addresses are allowed” or “these addresses are excluded” should not require maintaining a second list or duplicating state. It should simply be a property of how that policy is referenced.

By encoding inversion directly into the policy reference, the same registry entry becomes a reusable building block across standalone policies and Composite Policies. This makes expressions such as `A OR B`, `A AND NOT B`, or `NOT A` straightforward without introducing additional registries purely to represent the inverse of existing state.

stephancill (commented):
lgtm, defer to @stevieraykatz for final stamp

changelog/03_Denim_PolicyRegistry_not_policy.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Background
+
+### Policy Registry
+
+The Policy Registry is a singleton precompile at `0x8453000000000000000000000000000000000002`. B20 tokens call it for pre-operation compliance checks on an address.
+
+B20 stores a `uint64` policy ID per scope (`TRANSFER_FROM`, `TRANSFER_TO`, `MINT_RECEIVER`, `SEIZE_EXEMPT`, and other scopes) and calls `isAuthorized(policyId, account)` before gated operations.
+
+`isAuthorized` never reverts. A malformed or unknown ID returns `false` (deny).
```
stevieraykatz: this seems out of place? not sure why its included here.

rayyan224: Because we are using the isAuthroized function below since this change is on query time

changelog/03_Denim_PolicyRegistry_not_policy.md
```diff
@@ -0,0 +1,184 @@
...
+policyRegistry.createCompositePolicy(admin, INTERSECT, [kycId, notSanctions]);
+```
+
+Fail-closed: for any never-created base, `isAuthorized(base | INVERTED_POLICY_BIT, account) == false`. The inverted unknown ID never becomes allow-everyone.
+
+Round-trip: `invertedPolicyId(invertedPolicyId(id)) == id` (involutive). `policyExists(notSanctions) == policyExists(sanctionsId)`.
+
+## Design Decisions & Alternatives Considered
+
+Three options were weighed. The team converged on Option 2 (invert bit on the ID), implemented as Option 2a: Option 2 plus a base-existence guard that makes it fail-closed.
```
stevieraykatz: I find it a little strange we say "we landed on alternative 2" then list alts 1 and 3. I'd instead just rename then Alts 1 and 2 and not talk about what we landed on. They're explicitly alternatives and shouldn't reference pre-work

changelog/03_Denim_PolicyRegistry_not_policy.md
rayyan224: Folded the never-reverts sentence into the `isAuthorized` setup so it reads as invert's inherited query-time contract, not a stray fact.

changelog/03_Denim_PolicyRegistry_not_policy.md
rayyan224: Dropped the Option 2 / 2a preamble and the original-recommendation note. Rejected designs are now Alternative 1 and Alternative 2; fail-closed stays in the chosen-design paragraph.

stevieraykatz (approved):

PR #224, opened 2026-09-10
feat(policy): enforce TRANSFER_EXECUTOR_POLICY on every transfer path

rayyan224:
## Summary

Makes `TRANSFER_EXECUTOR_POLICY` apply to **every** transfer path. The executor gate now checks `msg.sender` on `transfer`, `transferFrom`, `transferWithMemo`, and `transferFromWithMemo` — including when `msg.sender == from`. Previously it ran only on the delegated `transferFrom` paths, and only when `msg.sender != from`.

This targets the Q4 "Denim" candidate *"Apply transfer executor policy on normal transfer"* (P2) — letting issuers use an executor allowlist to restrict **who may initiate a transfer** (e.g. only an approved settlement contract).

## Why

The old behavior left the executor scope unenforceable as an initiator gate, via two bypasses:

1. **Direct `transfer` was never gated** — the initiator is `msg.sender` (== `from`), and the check ran only inside `transferFrom`.
2. **Self-`transferFrom` skipped the check** — `msg.sender == from` bypassed it, so a non-allowlisted holder could route `transferFrom(self, to, amount)` to move tokens anyway.

Centralizing the check in `_transfer` on `msg.sender` and removing the `msg.sender == from` carve-out closes both.

## Approach

- **MockB20** (`test/lib/mocks/MockB20.sol`): executor check moved into `_transfer` (first, before sender/receiver, under the existing `_isPrivileged()` bootstrap bypass); duplicated body checks and the `msg.sender == from` carve-out removed. Allowance is still consumed in the `transferFrom*` bodies first, so revert order is unchanged.
- **Tests**: executor cases (sentinel / external allowlist / privileged bypass) on `transfer.t.sol` + memo parity; EXECUTOR woven into `transfer_revertOrder.t.sol` (C(7,2)=21 pairs) and the memo sequential order test; the old `transferFrom` self-caller skip test **inverted** into `test_transferFrom_revert_selfCaller_executorPolicyForbids` to pin the closed loophole.
- **Interface/comments**: `IB20.sol` natspec for the executor scope + the `transfer`/`transferFrom` revert lists; stale "delegated-only" comments in the mock/storage/revert-order headers.

> Scope note: docs (`docs/`) and changelog were intentionally left out of this PR.

## Compatibility

Purely behavioral — **no new selectors, events, errors, or storage**. An unset executor slot stays always-allow, so tokens that never configured the policy are unaffected. Factory bootstrap bypass and allowance accounting are unchanged.

**Breaking** only for a token that has set a restrictive `TRANSFER_EXECUTOR_POLICY` and relies on holders moving their own tokens via `transfer` / self-`transferFrom` — those holders must now be authorized as initiators.

## Testing

`forge test` — 746 passed, 0 failed, 4 skipped (pre-existing mock-only privileged skips).

🤖 Generated with [Claude Code](https://claude.ai/code)


test/lib/mocks/MockB20.sol
```diff
@@ -247,16 +237,10 @@ abstract contract MockB20 is IB20 {
...
         _requireNonZeroActors(from, to);
         // Allowance is consumed unconditionally — including during the factory
-        // bootstrap window — matching the Rust precompile.
-        // Only the executor-policy check below is bypassed for a privileged
-        // caller; infinite allowance is still not decremented.
+        // bootstrap window — matching the Rust precompile. Infinite allowance
+        // is still not decremented. The executor policy is enforced centrally
+        // in `_transfer` (on `msg.sender`), which honors the bootstrap bypass.
         _consumeAllowance(from, msg.sender, amount);
-        if (!_isPrivileged() && msg.sender != from) {
```
rayyan224, line 254: Removed here, becaues repulled in the helper no need for an extra SLOAD

changelog/03_Denim_B20_transfer_executor_enforcement.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Summary
+
+This change makes `TRANSFER_EXECUTOR_POLICY` apply to every transfer path. The executor gate now checks `msg.sender` on `transfer`, `transferFrom`, `transferWithMemo`, and `transferFromWithMemo`, including when `msg.sender == from`. Previously the check ran only on the delegated `transferFrom` paths, and only when `msg.sender != from`.
+
+The change is purely behavioral. It adds no new selectors, events, errors, or storage. A token that never sets `TRANSFER_EXECUTOR_POLICY` keeps the unset always-allow default, so it is unaffected.
+
+## Motivation
+
+An issuer may want an executor allowlist: only specific, approved contracts or accounts may initiate a transfer, for example a settlement contract that moves tokens on a holder's behalf. `TRANSFER_EXECUTOR_POLICY` exists for this, but the previous scope could not enforce it consistently with `TRANSFER_SENDER_POLICY` and `TRANSFER_RECEIVER_POLICY`, which already run on every transfer path.
```
rayyan224, line 16: Talk about transfer agent, here

changelog/03_Denim_B20_transfer_executor_enforcement.md
```diff
@@ -0,0 +1,163 @@
...
+- **Feature Name**: transfer_executor_enforcement
+- **Start Date**: 2026-09-10
+- **Authors**: Rayyan Alam
+- **Title**: (Breaking) Transfer Executor Policy Enforcement
+
+## Summary
+
+This change makes `TRANSFER_EXECUTOR_POLICY` apply to every transfer path. The executor gate now checks `msg.sender` on `transfer`, `transferFrom`, `transferWithMemo`, and `transferFromWithMemo`, including when `msg.sender == from`. Previously the check ran only on the delegated `transferFrom` paths, and only when `msg.sender != from`.
+
+The change is purely behavioral. It adds no new selectors, events, errors, or storage. A token that never sets `TRANSFER_EXECUTOR_POLICY` keeps the unset always-allow default, so it is unaffected.
```
rayyan224, line 12: Make sure to talk about the breaking change in the summary

changelog/03_Denim_B20_transfer_executor_enforcement.md
```diff
@@ -0,0 +1,163 @@
...
+        XE[executor policy] --> XS[sender policy]
+        XS --> XR[receiver policy]
+        XR --> XB[balance]
+    end
+```
+
+When more than one check would fail, the caller sees the first revert in that order:
+
+There are no new storage slots. The `TRANSFER_EXECUTOR_POLICY` policy ID is read from the same packed slot as before; `_transfer` now reads all three transfer-side policy IDs from that slot in one `SLOAD` instead of the executor lane being pre-warmed by a separate read in `transferFrom`'s body.
+
```
rayyan224, line 106: Add Gas Section

changelog/03_Denim_B20_transfer_executor_enforcement.md
```diff
@@ -0,0 +1,163 @@
...
+### Behavioural Changes
+
+The executor check moves from the `transferFrom` and `transferFromWithMemo` bodies into `_transfer`, where it runs first, before the existing sender and receiver checks, and under the same `_isPrivileged()` bootstrap bypass. The `msg.sender == from` carve-out that previously skipped the check is removed. `transfer` and `transferWithMemo` route through the same `_transfer` function, so they gain the check with no entrypoint-specific code.
+
+The previous order, by entrypoint:
+
+```mermaid
+flowchart TD
+    subgraph beforeTransfer ["Before: transfer / transferWithMemo"]
```
rayyan224, line 57: Check to move everything in the helper.

test/lib/mocks/MockB20.sol
```diff
@@ -754,18 +738,22 @@ abstract contract MockB20 is IB20 {
...
-            // One SLOAD pulls both policy IDs we need for the transfer
-            // check (and was already warmed if we came in via transferFrom,
-            // which reads the executor lane of the same slot first).
-            // Solidity emits a single SLOAD for the struct read + masked
-            // extracts for the named fields.
+            // One SLOAD pulls all three policy IDs we need for the transfer
+            // check. Solidity emits a single SLOAD for the struct read +
+            // masked extracts for the named fields.
             MockB20Storage.TransferPolicyIds memory packed = MockB20Storage.layout().transferPolicyIds;
+            if (!IPolicyRegistry(POLICY_REGISTRY).isAuthorized(packed.executor, msg.sender)) {
```
rayyan224, line 754: Add optimization for, cache

changelog/03_Denim_B20_transfer_executor_enforcement.md
rayyan224: Motivation now opens with the registered transfer-agent pattern: only the agent contract may initiate, via `transferFrom` against a holder allowance. The example uses `transferAgent` instead of `settlementContract`.

changelog/03_Denim_B20_transfer_executor_enforcement.md
rayyan224: Summary now states the breaking case up front: a token that already set a restrictive `TRANSFER_EXECUTOR_POLICY` now requires holders to be authorized initiators. Unset tokens stay always-allow.

changelog/03_Denim_B20_transfer_executor_enforcement.md
rayyan224: Added a `### Gas` section under Specs. It covers the single packed `SLOAD`, the extra `isAuthorized` on `transfer`/`transferWithMemo`, and that unset `ALWAYS_ALLOW_ID` still hits the registry.

changelog/03_Denim_B20_transfer_executor_enforcement.md
rayyan224: Checked: only the three transfer-side policy checks belong in `_transfer`. Pause stays a shared modifier; allowance is `transferFrom`-only and must run after zero-actor checks. Folding those in would change revert order or need an entrypoint flag. Documented as a rejected alternative.

test/lib/mocks/MockB20.sol
rayyan224: `_transfer` now caches the Policy Registry handle and the three unpacked policy IDs, since each ID is used twice (check + revert payload).

test/lib/mocks/MockB20.sol
```diff
@@ -754,23 +738,36 @@ abstract contract MockB20 is IB20 {
...
+            uint64 executorPolicy = packed.executor;
+            uint64 senderPolicy = packed.sender;
+            uint64 receiverPolicy = packed.receiver;
+            if (!registry.isAuthorized(executorPolicy, msg.sender)) {
+                revert PolicyForbids(TRANSFER_EXECUTOR_POLICY, executorPolicy);
+            }
+            // Same (policyId, account) as the executor check — skip the second
+            // registry call. Distinct IDs or a spender (`msg.sender != from`)
+            // still need both lookups.
+            bool skipSender = msg.sender == from && executorPolicy == senderPolicy;
```
rayyan224: Saves about 100 gas not much bot something 😅

ilikesymmetry: nit on variable name: `skipSenderPolicyCheck`

otherwise lgtm!

test/lib/mocks/MockB20.sol
```diff
@@ -754,23 +738,36 @@ abstract contract MockB20 is IB20 {
...
     function _transfer(address from, address to, uint256 amount) internal {
         if (!_isPrivileged()) {
-            // One SLOAD pulls both policy IDs we need for the transfer
-            // check (and was already warmed if we came in via transferFrom,
-            // which reads the executor lane of the same slot first).
-            // Solidity emits a single SLOAD for the struct read + masked
-            // extracts for the named fields.
+            // One SLOAD pulls all three policy IDs we need for the transfer
+            // check. Solidity emits a single SLOAD for the struct read +
+            // masked extracts for the named fields. Cache the registry handle
```
stevieraykatz: Not sure I like the framing of "Solidity emits". Emitting is specifically related to eventing. "Solidity compiles this to" seems more fitting?

test/lib/mocks/MockB20.sol
```diff
@@ -754,23 +738,36 @@ abstract contract MockB20 is IB20 {
...
-    ///      policy in their bodies before calling here; only the
-    ///      executor-policy check honors the bootstrap bypass,
-    ///      consistent with the sender/receiver policy bypass below.
+    ///      precompile) before calling here.
+    ///
+    ///      Enforces the executor (`msg.sender`), sender (`from`), and receiver
+    ///      (`to`) policies. Gating the executor here — not just on delegated
+    ///      `transferFrom` — lets an executor allowlist restrict who may
+    ///      initiate any transfer, including a holder moving their own tokens.
+    ///      All honor the bootstrap bypass; an unset lane is always-allow.
```
stevieraykatz: This is a weird place to land this natspec tbh. It's an internal so I wouldn't expect integrators to read here very often. I'd expect this kind of language as `@dev` blocks on the externals.

test/lib/mocks/MockB20.sol
rayyan224: Renamed to `skipSenderPolicyCheck`.

test/lib/mocks/MockB20.sol
rayyan224: Updated to "Solidity compiles this to a single SLOAD …" (same wording fix on `_writePolicyId`).

test/lib/mocks/MockB20.sol
rayyan224: Trimmed the product rationale off `_transfer`. Added `@dev` on `IB20.transfer` / `transferFrom` noting the executor check runs even when `msg.sender == from`.

stevieraykatz (approved):

PR #225, opened 2026-09-11
changelog(denim): add NOT / invert policy spec

rayyan224:
## Summary
- Registers **Denim** as hardfork ordinal `03` in `changelog/README.md` (next after Cobalt's `02`).
- Adds `changelog/03_Denim_PolicyRegistry_not_policy.md`: full spec for query-time policy inversion — a NOT flag in bit 63 of a `uint64` policy ID, the new `invertedPolicyId` view, `isAuthorized`/getter behavior on inverted IDs, composite-child invert handling, and the two rejected alternatives (new `NOT` policy type, per-child invert bitmask).
- Updates `docs/concepts/policies.md` with the corresponding "Inverting a policy" section (§2.3, renumbering everything after it by one).
- Not breaking: all existing selectors/events/errors are unchanged; existing policy IDs have bit 63 unset, so behavior is identical.

## Test plan
- [x] Docs-only change; no build/test impact.
- [ ] Reviewer check: filename/ordinal matches `changelog/AGENTS.md`'s naming rule, and the `docs/concepts/policies.md` renumbering (2.3 → 2.4, 2.3.1 → 2.4.1, etc.) is internally consistent.

### DIFF
--- a/changelog/03_Denim_PolicyRegistry_not_policy.md
+++ b/changelog/03_Denim_PolicyRegistry_not_policy.md
@@ -1,0 +1,183 @@
+# NOT / Invert Policies
+
+- **Feature Name**: not_policy
+- **Start Date**: 2026-09-09
+- **Authors**: Rayyan Alam
+- **Title**: NOT / Invert Policies
+
+## Summary
+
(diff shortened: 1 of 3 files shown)


rayyan224:
Superseded — folded into #222 since the docs describe the same NOT/invert-policy implementation that lives there.

PR #226, opened 2026-09-17
fix(smoke): update AbiDecodeFailed shape and drop stale burnBlocked ABI fragment

rayyan224:
## Summary
- `AbiDecodeFailed` reverts now carry a bare function selector with no message; relaxed `chain.py`'s length check (and docs) from requiring `> 4` bytes to `>= 4` bytes so the `factory` and `invariants` journeys stop failing on this expected shape change. The selector-match check (which actually verifies the error *type*) is unchanged.
- Removed the stale `_BURN_BLOCKED_FRAGMENT` / `_burn_blocked_at` workaround in `stablecoin_lifecycle.py`. It was added in #187 after #186 removed `burnBlocked` from `IB20`, but #193 later restored `burnBlocked` to the interface — so the hand-built fragment now duplicates the real ABI entry, which a stricter web3.py raises as a colliding-selector `Web3ValueError` on contract construction. Both call sites now bind through the normal `tok` contract object.

## Test plan
- [x] `PYTHONPATH=script python -m smoke factory -k` — passes, `AbiDecodeFailed` check holds
- [x] `PYTHONPATH=script python -m smoke invariants -k` — passes, `AbiDecodeFailed` check holds
- [x] `PYTHONPATH=script python -m smoke stablecoin -k` — passes, `burnBlocked` seize path works
- [x] `make smoke KEEP_GOING=1` (full suite) — 7 passed, 0 failed, 0 skipped


script/smoke/journeys/stablecoin_lifecycle.py
```diff
@@ -8,27 +8,9 @@
...
 from .. import config
-from ..abis import STABLECOIN_ABI
 from ..chain import Chain, log, step
 from ..codec import StablecoinCreateParams, init_call
 
-# `burnBlocked` is a deprecated precompile selector that base-std #186 removed from the IB20
-# interface, so it is absent from STABLECOIN_ABI. The selector still dispatches on the precompile
-# (retained for back-compat), so bind it via an explicit fragment to keep exercising the legacy
-# burn-based freeze-and-seize path. The transfer-based seize surface is covered by the `seize` journey.
-_BURN_BLOCKED_FRAGMENT = {
```
rayyan224, line 19: This broke the web3.py abi fragment since burnblocked was already added in

stephancill (approved):

